PR #11, opened 2015-10-14
Moved to Web Platform WG

plehegar:
New WG, new license.

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,15 +17,16 @@
                     company:    "The Paciello Group",
                     companyURL: "http://www.paciellogroup.com" }
           ],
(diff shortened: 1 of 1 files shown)


stevefaulkner:
will this effect auto publishing? from http://www.w3.org/TR/html-aria/

plehegar:
Steve, we may need to request help for the webmaster for the first one to go through. not sure.

stevefaulkner:
have updated so is same as PR https://github.com/w3c/html-aria/commit/2c7b98a002ad611bf18d86297c6b9330e7d90784

PR #109, opened 2018-05-29
fixed capitalization of "alertdialog"

geppy:
[ARIA's `alertdialog` term](https://www.w3.org/TR/wai-aria-1.1/#alertdialog) has no capital letters in it.


geppy:

stevefaulkner (approved):

Issue #135, opened 2019-05-21
Consider dropping invalid example iframe


marcoscaceres:
[Example 1](https://w3c.github.io/html-aria/#example-1) contains describes an abuse of ARIA and also contains an iframe to an invalid document... I wonder if it's worth dropping that iframe? 

Personally, I think the example is sufficiently clear of what's a bad thing to do.

PR #137, opened 2019-05-21
chore: drop details polyfill

marcoscaceres:
As of Edge 75, details is supported across all browsers. Probably safe to drop it. 

Also related to #135

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/137.html" title="Last updated on Jun 28, 2019, 6:00 AM UTC (f7d7699)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/137/137be38...f7d7699.html" title="Last updated on Jun 28, 2019, 6:00 AM UTC (f7d7699)">Diff</a>


stevefaulkner:
@marcoscaceres since we don't have a good idea of the browsers and versions that are used by our audience I would prefer to keep the pollyfill around for a while longer

PR #155, opened 2019-07-04
Emptyreferences

### LINKED ISSUE #145: HTML Standard inserted into empty links
In index.html the table, ARIA Roles, States and Properties, has four instances of empty `<a>` elements.  Previously these were not shown to users but with recent changes they now all show a link labelled "HTML Standard". The empty links can be found in the following rows:

- button Descendant restrictions`<a data-cite="html/dom.html#phrasing-content-2"></a>`
- checkbox Descendant restrictions `<a data-cite="html/dom.html#phrasing-content-2"></a>`
- dialog Kind of content `<a data-cite="html/dom.html#interactive-content"></a>`
- heading Descendant restrictions `<a data-cite="html/dom.html#phrasing-content-2"></a>`

Should each of these be given a proper name or should they be removed?

pgrucza:
removes empty references that were not displayed previously
fixes #145

***
<a href="https://pr-preview.s3.amazonaws.com/pgrucza/html-aria/pull/155.html" title="Last updated on Jul 4, 2019, 7:50 PM UTC (2f7c7ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/155/1e37a84...pgrucza:2f7c7ff.html" title="Last updated on Jul 4, 2019, 7:50 PM UTC (2f7c7ff)">Diff</a>


index.html
```diff
@@ -2895,7 +2895,7 @@ <h2>
             </td>
             <td>
               <a data-cite="html/dom.html#flow-content">Flow
-              content</a><a data-cite="html/dom.html#phrasing-content-2"></a>,
+              content</a>,
```
scottaohara, line 2898: maybe this is more the result of a partial merge, rather than an errant empty link?

for instance, being in the restricted column, flow content is not allowed as a child of button, nor is phrasing content that is interactive (which could be what the intent was of the link removed?)

it's a bit hard to distinguish because the expected child content of a button should be:
>Phrasing content, but there must be no interactive content descendant.

pgrucza: Yes, for an HTML button element that is true but is the same rule applied when we assign a role="button" to an element? 
Currently the Nu Html Checker provides two different results for the following:

- `<button type="button"><h1>hello</h1></button>` This is flagged as an error
- `<div role="button"><h1>hello</h1></div>` This is not flagged as an error

I was using the link and menuitem roles for comparison but that may not be the best approach.

scottaohara: Seems like an oversight of the validator, rather than an example of why it's OK.

pgrucza: Beyond button, this request also removes the empty links found in checkbox, dialog and heading. Is there a problem with removing these from those 3 roles?

stevefaulkner: I reviewed and it looks like these empty links were included in my initial commit https://github.com/w3c/html-aria/commit/75f41b64ed16308e56da5e85cad8f7d5820c8c35, they are not supposed to be there so fine to remove

stevefaulkner: @pgrucza @scottaohara is correct in that if an element with an implicit aria role permits only certain types of content, this should be generally reflected in the permissions for the role, although it is not one to one as the reason why people use aria roles rather than native elements, sometimes, is so they can do stuff that results in better styling, in the example case it is not a good idea either way to have a heading nested in a button, so the restrictions need to be tightened.

scottaohara (approved):
Per steve's comments, we should pull this in and then take another look at these restrictions to make sure they're up to date.

scottaohara:

PR #159, opened 2019-08-02
fix typo for `application`

yuheiy:
- Remove the changed marker in Description cell
- Remove `aria-expanded` from Supported Properties cell

***
<a href="https://pr-preview.s3.amazonaws.com/yuheiy/html-aria/pull/159.html" title="Last updated on Aug 2, 2019, 8:02 AM UTC (27ccf2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/159/da01b87...yuheiy:27ccf2e.html" title="Last updated on Aug 2, 2019, 8:02 AM UTC (27ccf2e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2797,7 +2797,6 @@
               A structure containing one or more focusable elements requiring
               user input, such as keyboard or gesture events, that do not
               follow a standard interaction pattern supported by a widget role.
-              - <span class="changed-feature">changed)</span>
             </td>
             <td>
               none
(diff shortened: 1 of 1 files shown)


scottaohara:
Thanks for filing the issue, @yuheiy 

There does seem to be a bit of a misalignment here.

[aria 1.1](https://www.w3.org/TR/wai-aria-1.1/#application) notes that `aria-expanded` is not supported on `role=application`.  However, [aria 1.2](https://w3c.github.io/aria/#application) has it listed under the supported states and properties.

There is an issue that was just opened to [clarify the allowed roles for `aria-expanded`](https://github.com/w3c/aria/issues/1026), where this specific allowance was brought up.  So for now it makes sense to wait on the outcome of that issue.

siusin:

scottaohara:
So, ARIA _will_ be allowing `aria-expanded` on `application`, at least in 1.2 but I don't think this will be changed going forward either.  

The changed marker has also been removed (along with all changed / new markers) via another PR.  So this PR is no longer necessary.  

Thank you for raising it / helping to prompt some further discussion over in the ARIA WG per this attribute allowance.

PR #189, opened 2019-10-18
define No role

### LINKED ISSUE #188: Define "No role"
[Section 2](https://w3c.github.io/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html) contains a table listing the HTML language feature, the implicit ARIA semantics for that feature, and the ARIA roles, states, and properties which may be used for that HTML language feature.

In a case where specific-but-limited roles may be used, they are listed. Each one is also linked to another location in the document that provides additional information about that role.

In a case where literally any valid role can be used, the string `Any Role` is provided and that text is linked to an in-page anchor stating "Where a cell in the third column includes the term Any role it indicates that any role value apart from the implicit ARIA semantics role value, MAY be used."

However, in a case where the HTML language feature *should not* be given a role, the string `No Role` is used.  However, unlike the prior case, there is no hyperlink to any definition of what this means.

While it might appear to be prima facie obvious what "No role" means, I think authors would benefit from a definition, similar in spirit to the definition provided for "Any Role"

Suggested wording:

> "Where a cell in the third column includes the term No role it indicates that authors SHOULD NOT attempt to override native semantics by providing a `role` for this element."

scottaohara:
closes #188 

@stevefaulkner please review / word smith as necessary.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/189.html" title="Last updated on Oct 18, 2019, 4:34 PM UTC (eb1197e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/189/c18ec93...eb1197e.html" title="Last updated on Oct 18, 2019, 4:34 PM UTC (eb1197e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -107,7 +107,7 @@
         used. Where a cell in the third column includes the term
         <dfn><strong>Any</strong> `role`</dfn> it indicates that any `role`
         value apart from the <a>implicit ARIA semantics</a> `role` value, MAY
-        be used. If an cell in the third column includes the term <dfn><strong>No `role`</strong></dfn> it indicates that authors SHOULD NOT overwrite the implicit ARIA semantics, or native semantics of the HTML element.
+        be used. If an cell in the third column includes the term <dfn><strong>No `role`</strong></dfn> it indicates that authors MUST NOT overwrite the implicit ARIA semantics, or native semantics of the HTML element.
       </p>
 
       <p class="note" id="aria-usage-note">
--- a/index.html
+++ b/index.html
@@ -97,16 +97,19 @@
         requirements for the use of ARIA markup in HTML documents and describes
         the <a>implicit ARIA semantics</a> that apply to <a data-cite=
         "html/dom.html#elements">HTML elements</a> as defined in the HTML
-        Accessibility API Mappings 1.0 [[html-aam-1.0]] specification. Each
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -97,16 +97,19 @@ <h2 id="docconformance">
...
+      <p>
+        Each language feature (element or attribute) in a cell in the first column
         implies the ARIA semantics (any role, states, and properties) given in
         the cell in the second column of the same row. The third cell in each
         row defines which ARIA role values and aria-* attributes which MAY be
         used. Where a cell in the third column includes the term
         <dfn><strong>Any</strong> `role`</dfn> it indicates that any `role`
         value apart from the <a>implicit ARIA semantics</a> `role` value, MAY
-        be used.
+        be used. If an cell in the third column includes the term <dfn><strong>No `role`</strong></dfn> it indicates that authors SHOULD NOT overwrite the implicit ARIA semantics, or native semantics of the HTML element.
```
stevefaulkner: MUST NOT

stevefaulkner: more simply it means that the role attribute is not allowed on the element

scottaohara: MUST it is :)

stevefaulkner (approved):

PR #191, opened 2019-11-18
allow aria-setsize and aria-posinset on tr in treegrid

### LINKED ISSUE #190: aria-setsize and aria-posinset are allowed on tr in treegrid
Please see https://github.com/w3c/aria/issues/558

aria-setsize and aria-posinset were allowed on tr (when descendant of treegrid) in the ARIA spec in August 2018, but it seems that propagating that change to ARIA in HTML must have slipped through the cracks.

[ARIA for role="row"](http://w3c.github.io/aria/#row) currently says:
> While the row role can be used in a table, grid, or treegrid, the semantics of aria-expanded, aria-posinset, aria-setsize, and aria-level are only applicable to the hierarchical structure of an interactive tree grid. Therefore, authors MUST NOT apply aria-expanded, aria-posinset, aria-setsize, and aria-level to a row that descends from a table or grid, and user agents SHOULD NOT expose any of these four properties to assistive technologies unless the row descends from a treegrid.

So ARIA in HTML needs to be updated as follows:
- [tr element](http://w3c.github.io/html-aria/#el-tr) needs to allow explicit `role="row"` if in "treegrid" (as well as grid), i.e.:
    > role=row, may be explicitly declared when child of a table element with role=grid ** or treegrid**
- [row role](http://w3c.github.io/html-aria/#index-aria-row) needs to add aria-setsize and aria-posinset to the list of supported properties, and (somehow) say that aria-expanded, aria-posinset, aria-setsize, and aria-level are only supported if in a treegrid

I will provide a PR shortly.

carmacleod:
- resolves #190

Note: Not quite sure how you want to add the slightly complicated semantics of some row properties only being allowed in context of a treegrid. I put everything into the same "Supported properties" cell (because it's more compact), but maybe you prefer splitting it into 2 rows (one for `role="row" if child of treegrid` and another for `role="row" if not child of treegrid`)?

***
<a href="https://pr-preview.s3.amazonaws.com/carmacleod/html-aria/pull/191.html" title="Last updated on Nov 21, 2019, 11:57 PM UTC (82425f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/191/010683d...carmacleod:82425f4.html" title="Last updated on Nov 21, 2019, 11:57 PM UTC (82425f4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2290,7 +2290,7 @@
             <td>
               <a href="#index-aria-row">`role=row`</a>, may be
               explicitly declared when child of a `table` element
-              with `role=grid`
+              with `role=grid` or `treegrid`
             </td>
             <td>
               <p>
@@ -4082,6 +4082,7 @@
               none
             </td>
             <td>
+              <p>If child of `role=grid`, `rowgroup`, `table` or `treegrid`:</p>
               <ul>
                 <li>
                   <a data-cite="wai-aria-1.1#aria-colindex">`aria-colindex`</a>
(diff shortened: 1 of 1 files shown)


stevefaulkner (approved):

scottaohara (approved):

scottaohara:

PR #210, opened 2020-02-03
updates to custom elements

### LINKED ISSUE #180: update custom element
Per: [HTML issue 4658](https://github.com/whatwg/html/pull/4658) and [ARIA issue #984](https://github.com/w3c/aria/pull/984) it should be noted that a custom element may have no corresponding role (map to generic) or may have a role set by `ElementInternals`. 

If the role is set in that way, and the custom element doesn't merely act as a container for child elements (light or shadow dom) with their own roles, then the listing for the custom element would need to indicate that it either maps to generic, OR the role set by `ElementInternals`.  

In that case, the allowed state/properties/naming/etc. would be dependent on the role the element represents.

scottaohara:
Closes #180 

Updates autonomous custom elements and form associated custom elements to indicate that they have no corresponding role if an author did not define a role in the custom element’s `ElementInternals`.

The allowed roles and properties column has been updated to indicate that if a custom element has an author defined role, then the allowed roles to that element must adhere to the defined role of the custom element.

I'm thinking it's probably a good idea to add an example, at the end of section 2, to help demonstrate proper conformance here.

e.g.
`<my-element>` with no role defined from its `ElementInternals` can accept any role.

but if `<my-check>` were created and had its role defined via its internals...
```js
class MyCheckbox extends HTMLElement {
  static get formAssociated() { return true; }
  static get observedAttributes() { return ['checked']; }

  constructor() {
    super();
    this._internals = this.attachInternals();
    this.addEventListener('click', this._onClick.bind(this));

    this._internals.role = 'checkbox';
    this._internals.ariaChecked = false;
  }
  // ...
```
(example from the [pending PR to the HTML spec](https://whatpr.org/html/4658/custom-elements.html#custom-elements-accessibility-example))
Then only roles allowed on a checkbox would be allowed on this custom element.

That all said, we shouldn't merge this update until the linked HTML and ARIA issues in #180 are merged.

cc @carmacleod

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/210.html" title="Last updated on Feb 13, 2021, 9:02 PM UTC (5b3d26f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/210/123406d...5b3d26f.html" title="Last updated on Feb 13, 2021, 9:02 PM UTC (5b3d26f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -946,7 +946,7 @@
                 <a href="#index-aria-switch">`switch`</a>
                 or <a href="#index-aria-textbox">`textbox`</a>.
               </p>
-              <p>Otherwise only allowed roles to the role exposed by the custom element's `ElementInternals`.</p>
+              <p>Otherwise only allowed roles for the role exposed by the custom element's `ElementInternals`.</p>
               <p>
                 <a href="#index-aria-global">Global `aria-*` attributes</a> and
                 any `aria-*` attributes applicable to the allowed roles and
--- a/index.html
+++ b/index.html
@@ -337,12 +337,14 @@
               <a>autonomous custom element</a>
             </td>
             <td>
-              <a>No corresponding role</a>
+              Role exposed from author defined <a href="https://html.spec.whatwg.org/multipage/custom-elements.html#the-elementinternals-interface">`ElementInternals`</a>.
+              Otherwise <a>no corresponding role</a>.
             </td>
             <td>
               <p>
-                <a><strong>Any</strong> `role`</a>
+                If no author defined role,<br><a><strong>Any</strong> `role`</a>
               </p>
+              <p>Otherwise only allowed roles to the role exposed by the custom element's `ElementInternals`.</p>
               <p>
                 <a href="#index-aria-global">Global `aria-*` attributes</a> and
                 any `aria-*` attributes applicable to the allowed roles and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -337,12 +337,14 @@ <h2 id="docconformance">
...
-              <a>No corresponding role</a>
+              Role exposed from author defined <a href="https://html.spec.whatwg.org/multipage/custom-elements.html#the-elementinternals-interface">`ElementInternals`</a>.
+              Otherwise <a>no corresponding role</a>.
             </td>
             <td>
               <p>
-                <a><strong>Any</strong> `role`</a>
+                If no author defined role,<br><a><strong>Any</strong> `role`</a>
               </p>
+              <p>Otherwise only allowed roles to the role exposed by the custom element's `ElementInternals`.</p>
```
carmacleod:
```suggestion
              <p>Otherwise only allowed roles for the role exposed by the custom element's `ElementInternals`.</p>
```

I think (?) "for" flows better here?

index.html
```diff
@@ -944,6 +946,7 @@ <h2 id="docconformance">
                 <a href="#index-aria-switch">`switch`</a>
                 or <a href="#index-aria-textbox">`textbox`</a>.
               </p>
+              <p>Otherwise only allowed roles to the role exposed by the custom element's `ElementInternals`.</p>
```
carmacleod:
```suggestion
              <p>Otherwise only allowed roles for the role exposed by the custom element's `ElementInternals`.</p>
```

carmacleod:
Looks good.
Agree that an example would be really useful in this case.
How does this document refer to examples in section 2?
i.e. For the [fish example](https://pr-preview.s3.amazonaws.com/w3c/html-aria/210/46fece9...5388b6c.html#example-2), I would have thought that the [role `img` table row](https://pr-preview.s3.amazonaws.com/w3c/html-aria/210/46fece9...5388b6c.html#index-aria-img) might refer to it (e.g. "See EXAMPLE 2"), but I don't see any reference?

scottaohara:
yes, no references presently.  we should make them

mtrootyy:
https://github.com/whatwg/html/pull/4658 was merged.

carmacleod:
Also https://github.com/w3c/aria/pull/984 was merged. So I think the "not ready for merge" label can be removed now.

PR #212, opened 2020-02-15
limit the allowed roles on img elements

### LINKED ISSUE #166: Disallow role of textbox on img elements
The [ARIA in HTML specification](https://www.w3.org/TR/html-aria/#img-alt) already imposes this limitation on `img`-elements without an `alt`-attribute. Unless someone can come up with a good example of where this would be useful, I think it makes sense to disallow the `textbox`-role from being used on `img`-elements with an `alt`-attribute, too.

As an example of potential usability, @stevefaulkner [mentioned a read-only textbox](https://github.com/w3c/aria/issues/711#issuecomment-535035780). If that were indeed a valid (and valuable) situation to cater to it might be worth requiring [`aria-readonly`](http://w3c.github.io/aria/#aria-readonly) when setting the `textbox` role on an `img`.

scottaohara:
closes #166

this PR does the following:

* limits the allowed roles on an `img` with `alt=some_text`.
  - button
  - link
  - menuitem
  - menuitemcheckbox
  - menuitemradio
  - option
  - radio
  - switch
  - tab
* disallows any role on an `img` without an `alt` attribute (some screen readers will ignore an `img` without an `alt`.  some will announce the file name.  rather we not allow a `role=button` without an accessible name, or with a name "photo-1.jpg").  

  If an `img` is provided an authored defined accessible name via other means (`aria-label`, `aria-labelledby`, `title`) then same allowances as `img alt="some text"`.

I'd be more than open to whittling the allowed roles down further, or hearing reasons to add some back in.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/212.html" title="Last updated on Feb 15, 2020, 11:37 AM UTC (dfb3259)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/212/9599ae9...dfb3259.html" title="Last updated on Feb 15, 2020, 11:37 AM UTC (dfb3259)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1108,7 +1108,7 @@
           </tr>
           <tr id="el-img" tabindex="-1">
             <td>
-              <code><a>img</a> with <a data-cite=
+              <code><a>img</a></code> with <code><a data-cite=
               "html/embedded-content.html#attr-img-alt">alt</a>="some text"</code>
             </td>
             <td>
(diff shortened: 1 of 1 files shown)


stevefaulkner (approved):
LGTM

rdeltour:
I believe `doc-cover` is missing in the list of allowed roles. Also, if `link` is allowed, then all the DPUB roles extending link should probably be allowed too.

scottaohara:
@rdeltour, PRs welcome :)

PR #224, opened 2020-03-20
add note about using “role=table” on table element

### LINKED ISSUE #223: Allow role="table" on <table> elements to enforce treatment as a data table
[Document conformance requirements for use of ARIA attributes in HTML](https://w3c.github.io/html-aria/#docconformance) specifies that the table element has an implicit ARIA role of table and thus that role="table" should not be used on this element. Specifically, the second column is labelled "Implicit ARIA semantics - SHOULD NOT be used" and the table element specifies role=table in this second column. While this generally makes sense, Firefox and Chrome both use heuristics to guess whether an HTML table is a layout table and provide that as a hint to clients, but there are edge cases where they get it wrong, particularly on very small tables. role="table" is a useful way of overriding this to force treatment as a data table, and this trick works in both Firefox and Chrome. As currently specified, this useful trick is a conformance violation. My hope is that we can tweak the spec to allow for this.

This is difficult to do cleanly because of the "SHOULD NOT be used " wording in the label of that second column, which applies to all of the rows. Perhaps we could:

1. Drop the "SHOULD NOT be used" wording. Alternatively, we could change it to something like "Usually should not be used", but that feels a bit fluffy.
2. Expand the note above to explain there are some specific cases where making implicit semantics explicit is acceptable. The note currently states:

    > Setting an ARIA role and/or aria-* attribute that matches the implicit ARIA semantics is unnecessary and is NOT RECOMMENDED as these properties are already set by the browser.

    Interestingly, "is not recommended" already seems weaker than the "Should not be used" in the table. Possible new wording:

    > In most cases, setting an ARIA role and/or aria-* attribute that matches the implicit ARIA semantics is unnecessary and should not be done, as these properties are already set by the browser. However, there is one case where this may be necessary, as noted in the table below.

3. Add a footnote in the second column for the table element explaining this use case. Possible wording:

> The HTML table element is sometimes used for layout tables. Therefore, browsers use heuristics to guess whether a table is a layout table and provide that as a hint to assistive technology products. In some cases, this guess can be incorrect, particularly on small tables. role="table" can be used to override this guess and force a table to be treated as a data table.

scottaohara:
closes #223

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/224.html" title="Last updated on Jul 28, 2020, 9:36 AM UTC (1b295ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/224/fd0a72e...1b295ce.html" title="Last updated on Jul 28, 2020, 9:36 AM UTC (1b295ce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2453,6 +2453,8 @@
             </td>
             <td>
               <code>role=<a href="#index-aria-table">table</a></code>
+
+              <p><strong>Note:</strong> some user agents suppress a table's <a>implicit ARIA semantics</a> if a table does not contain `th` elements, or is not provided a name by a `caption`, for example. Authors can use `role=table` to reinstate the role if necessary.
             </td>
             <td>
               <p>
--- a/index.html
+++ b/index.html
@@ -2454,7 +2454,7 @@
             <td>
               <code>role=<a href="#index-aria-table">table</a></code>
 
-              <p><strong>Note:</strong> some user agents suppress a table's <a>implicit ARIA semantics</a> if a table does not contain `th` elements, or is not provided a name by a `caption`, for example. Authors can use `role=table` to reinstate the role if necessary.
+              <p><strong>Note:</strong> some user agents suppress a table's <a>implicit ARIA semantics</a> if a table does not contain `th` elements, is not provided a name by a `caption`, or has its CSS `display` property altered, for example. Authors can use `role=table` to reinstate the role if necessary.
             </td>
             <td>
               <p>


scottaohara:
@marcoscaceres seems something screwy is going on with previews per the "Error: 403 Forbidden" 


never mind...fixed?

index.html
```diff
@@ -2453,6 +2453,8 @@ <h2 id="docconformance">
             </td>
             <td>
               <code>role=<a href="#index-aria-table">table</a></code>
+
+              <p><strong>Note:</strong> some user agents suppress a table's <a>implicit ARIA semantics</a> if a table does not contain `th` elements, or is not provided a name by a `caption`, for example. Authors can use `role=table` to reinstate the role if necessary.
```
```suggestion
              <p><strong>Note:</strong> some user agents suppress a table's <a>implicit ARIA semantics</a> if a table does not contain `th` elements, is not provided a name by a `caption`, or has its CSS `display` property altered, for example. Authors can use `role=table` to reinstate the role if necessary.
```

scottaohara:
rather than call out a bunch of different instances where doubling up on roles may be beneficial due to implementation quirks / heuristic magic, #237 will be used instead.

closing this pr

PR #235, opened 2020-07-13
Use isContentEditable rather than contenteditable

AlexLloyd0:
1. isContentEditable covers the case when the contenteditable property is inherited from an ancestor
2. "Element with contenteditable attribute" sounds like it also covers `contenteditable="false"` or `contenteditable="inherit"`, but it doesn't make sense that `aria-readonly="false"` in this case.

***
<a href="https://pr-preview.s3.amazonaws.com/AlexLloyd0/html-aria/pull/235.html" title="Last updated on Jul 14, 2020, 11:40 AM UTC (b89980d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/235/09f9eab...AlexLloyd0:b89980d.html" title="Last updated on Jul 14, 2020, 11:40 AM UTC (b89980d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2881,15 +2881,15 @@
           <tr id="att-contenteditable" tabindex="-1">
             <td>
               Element with <a data-cite=
-              "html/interaction.html#attr-contenteditable">`contenteditable`</a>
-              attribute
+              "html/interaction.html#attr-contenteditable">`isContentEditable`</a>
+              ="true"
             </td>
             <td>
               `aria-readonly="false"`
             </td>
             <td>
               Do not set `aria-readonly="true"` on an element that
-              has a `contenteditable` attribute set.
+              has `isContentEditable="true"`.
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -1534,7 +1534,7 @@
               <p>
                 Roles:
                 <a href="#index-aria-combobox">`combobox`</a>,
-                <a href="#index-aria-searchbox">`searchbox`</a></code>
+                <a href="#index-aria-searchbox">`searchbox`</a>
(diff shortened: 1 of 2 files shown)


AlexLloyd0:
I've now linked my github and W3 accounts

index.html
```diff
@@ -2881,15 +2881,15 @@ <h2 id="docconformance">
           <tr id="att-contenteditable" tabindex="-1">
             <td>
               Element with <a data-cite=
-              "html/interaction.html#attr-contenteditable">`contenteditable`</a>
-              attribute
+              "html/interaction.html#attr-contenteditable">`isContentEditable`</a>
```
scottaohara: agree with the update to add clarity, but would rather this be added as a description. Otherwise this would be the only instance of directly referring to an IDL attribute.

```suggestion
              "html/interaction.html#attr-contenteditable">`contenteditable`</a> where `isContentEditable="true"`
```

AlexLloyd0: The text now reads

> Element with `contenteditable` where `isContentEditable="true"`.

This sounds like only `first` should have `aria-readonly="false"` in the following example, but `contenteditable="inherit"` is the default so `first` and `second` should be identical.

```html
<div contenteditable="true">
  <div id="first" contenteditable="inherit"></div>
  <div id="second"></div>
<div>
```

I think the IDL attribute represents the correct semantics, an alternative to directly referring to it would be to describe its semantics, something like

> Element with `contenteditable="true"`; or
> Element without a `contenteditable` attribute whose closest ancestor with a `contenteditable` attribute has `contenteditable="true"`

What do you think?

scottaohara: I like your new proposed text.  Let's go with that.

scottaohara (approved):

Issue #228, opened 2020-05-13
`img` element says `role=img` should not be used


dd8:
Given the long-standing problem with VoiceOver / Safari which requires role=img on img elements this is potentially harmful.

For example, validator.w3.org/nu warns about this:

```
<img src='file.svg' alt='Image description' role='img' >
```
 
Fixing the warning makes the image less accessible to VO users.

Yes, VoiceOver / Safari shouldn't be doing this, but it has for at least 3 years and Apple seem resistant to changing this.

LJWatson:
The problem with making a change like this to a specification is that it
enables the browser and/or AT vendor to say they no longer have a bug in
their software, and that forever condemns everyone to implementing a
hack for the sake of a single implementation.

scottaohara:
agreed.  and upon further reflection I'm concerned we made the wrong choice in moving forward with #165 

changing specifications to cater to a single implementor doing things differently, or not fixing long standing bugs is a very slippery slope and I think we need to make sure we don't fall further down it.

dd8:
I'm not suggesting changing the browser/AT requirements in html-aam.

Apple obviously disagree with the mapping in html-aam spec, so are unlikely to fix this:
https://lists.w3.org/Archives/Public/wai-xtech/2014Feb/0024.html

I'm suggesting a change here in the document conformance requirements in html-aria for authors to remove "should not use role=img on img" - that's the bit that's causing harm.

LJWatson:
I'm not convinced it's the HTML mapping that is the problem. If VoiceOver announces "group" for an `<img>` with an SVG source file, it suggests that the old role for the `<svg>` element is being announced. If that's the case, it's the SVG mapping that is incorrect in WebKit/Safari.

dd8:
The VO SVG img issue pre-dates the introduction of the new SVG roles by several years:

- Steve F's email from 2014 described test results for Safari 6 which was released around July 2012.
- The first public draft of the WAI-ARIA Graphics Module was December 2015, so I don't think the new SVG roles were available when this was originally implemented

VO should use the new SVG roles, but that's a different issue.

Looking at the email from James Craig what they were trying to do was expose accessibility info from SVG referenced by an `img src` in the same way as inline `<svg>` elements. That seems more important with a touch interface (e.g. read out the number from this graph under my finger) which might explain their reluctance to change this.

I'll do some digging in the WebKit source to figure out exactly what the code does.

scottaohara:
[...]

PR #237, opened 2020-07-25
Softening language around explicitly assigning implicit ARIA semantics

### LINKED ISSUE #236: Softening the language around unnecessary explicit role?
As this comes up every now and again in discussions: currently, the advice given here in the rules https://w3c.github.io/html-aria/#rules-wd

> Web developers SHOULD NOT set the ARIA role and aria-* attributes to values that match the implicit ARIA semantics defined in the table.

and later in the conformance requirement note https://w3c.github.io/html-aria/#docconformance

> Setting an ARIA role and/or aria-* attribute that matches the implicit ARIA semantics is unnecessary and is NOT RECOMMENDED as these properties are already set by the browser.

SHOULD NOT / NOT RECOMMENDED may be a bit too strong here. per the RFC

> SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

But if authors simply explicitly define the same `role="..."` that is already implicit, there are no actual implications, right? There is no nasty/potential side effect that they need to be aware of...it's just the potential for them to get it wrong (which is a danger for any technology) and the fact that it's more verbose/unnecessary, but ultimately there is no possible harm that can come of it?

The only situation where I could imagine it potentially causing a problem is if you have an element that has different roles depending on the context it's used in (e.g. `header` that exposes different roles depending on whether or not it's in an `article` etc), but even here it's more a problem of "authors could get it wrong".

Would it perhaps make sense to drop the RFC "SHOULD NOT" / "NOT RECOMMENDED", and instead soften this into an explanation/note that says, more or less, that authors are encouraged not to explicitly use `role="..."` to just define the same implicit role the element already has, but that it's mainly because it's unnecessary? And not because it causes actual issues? Unless there are issues that I'm currently unaware of...

patrickhlauke:
Closes #236

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/237.html" title="Last updated on Aug 10, 2020, 8:46 PM UTC (a78d2c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/237/be57bdf...a78d2c1.html" title="Last updated on Aug 10, 2020, 8:46 PM UTC (a78d2c1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -90,8 +90,19 @@
       </aside>
 
       <aside class="example">
+        The following uses a `role=button` on a `button` element. This is unnecessary, this
+        is alreadt the implicit role the is exposed for the element. However,
+        in practice this will likely not have any unforeseen side effects other than unnecessarily
+        making the markup more verbose.
+        <pre class="HTML">
+          &lt;button role="button"&gt;...&lt;/button&gt;
+        </pre>
+      </aside>
+      
+      <aside class="example">
         The following uses a `role=group` on a `fieldset` element. This is unnecessary,
-        because the `fieldset` element is implicitly exposed as a `role=group`.
+        because the `fieldset` element is implicitly exposed as a `role=group`. Again,
+        in practice this will likely not have any unforeseen side effects.
         <pre class="HTML">
           &lt;fieldset role="group"&gt;...&lt;/fieldset&gt;
         </pre>
@@ -106,7 +117,7 @@
       </aside>
 
       <aside class="example">
-        The following uses a `role=button` on a `summary` element. This is unnecessary and can cause cross-platform issues. Setting `role=button` to the `summary` element can actually prevent correctly exposing this element to assistive technologies.
+        The following uses a `role=button` on a `summary` element. This is unnecessary and can cause cross-platform issues. Setting `role=button` to the `summary` element can actually this element from being correctly exposed to assistive technologies.
         <pre class="HTML">
           &lt;details>
             &lt;summary role="button"&gt;more information&lt;/summary&gt;
@@ -140,7 +151,7 @@
 
       <p class="note" id="aria-usage-note">
         While setting an ARIA `role` and/or `aria-*` attribute that matches the
-        <span>implicit ARIA semantics</span> is NOT RECOMMENDED, in some situations explicitly setting these attributes can be helpful. For instance, if supporting user agents which lack specific implicit semantics.
+        <span>implicit ARIA semantics</span> is NOT RECOMMENDED, in some situations explicitly setting these attributes can be helpful. For instance, in user agents which lack specific implicit semantics.
       </p>
 
       <table class="simple">
@@ -1201,7 +1212,7 @@
             </td>
             <td>
               <p>
-                If not provided an author defined accessible name by other methods:
+                If no author-defined accessible name is provided by other methods:
                 <strong class="nosupport">No `role`</strong>, and <strong>no `aria-*` attributes</strong> except
                 `aria-hidden`.
               </p>
@@ -2616,7 +2627,7 @@
                 descendant of a `table` element.
               </p>
               <!--<p><code>role=<a href="#index-aria-cell">gridcell</a></code> if a descendant
-                of a <code>table</code> elementwith <code>role=grid</code></p>-->
+                of a <code>table</code> element with <code>role=grid</code></p>-->
             </td>
             <td>
               <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -107,13 +110,19 @@ <h2 id="docconformance">
...
       </p>
 
       <p class="note" id="aria-usage-note">
         Setting an ARIA `role` and/or `aria-*` attribute that matches the
-        <span>implicit ARIA semantics</span> is unnecessary and is NOT
-        RECOMMENDED as these properties are already set by the browser.
+        <span>implicit ARIA semantics</span> is NOT RECOMMENDED. It is unnecessary,
+        as these properties are already set by the browser. However, explicitly
+        setting these attributes won't generally have an adverse effect (provided
+        that they are correct and indeed match the implicit semantics) - and this
```
carmacleod: Prefer a comma here?

```suggestion
        that they are correct and indeed match the implicit semantics), and this
```

index.html
```diff
@@ -87,6 +88,33 @@ <h2 id="rules-wd">
...
+      <aside class="example">
+        The following uses a `role=list` on an `ul` element. This is generally unnecessary,
+        because the `ul` element is implicitly exposed as a `role=list`. However, some user agents suppress a list's implicit ARIA semantics if list markers are removed. Authors can use `role=list` to reinstate the role if necessary, though this practice would generally not be recommended.
+        <pre class="HTML">
+          &lt;ul role="list"&gt;...&lt;/ul&gt;
+        </pre>
+      </aside>
+
+      <aside class="example">
+        The following uses a `role=button` on a `summary` element. This is unnecessary and can cause cross-platform issues. Setting `role=button` to the `summary` element can actually prevent correctly exposing this element to assistive technologies.
```
patrickhlauke: "prevent correctly exposing this element" > "prevent this element from being correctly exposed"

index.html
```diff
@@ -96,24 +124,23 @@ <h2 id="docconformance">
...
+        includes the term <dfn><strong>No `role`</strong></dfn> it indicates that authors MUST NOT
+        overwrite the implicit ARIA semantics, or native semantics of the HTML element.
       </p>
 
       <p class="note" id="aria-usage-note">
-        Setting an ARIA `role` and/or `aria-*` attribute that matches the
-        <span>implicit ARIA semantics</span> is unnecessary and is NOT
-        RECOMMENDED as these properties are already set by the browser.
+        While setting an ARIA `role` and/or `aria-*` attribute that matches the
+        <span>implicit ARIA semantics</span> is NOT RECOMMENDED, in some situations explicitly setting these attributes can be helpful. For instance, if supporting user agents which lack specific implicit semantics.
```
patrickhlauke: "if supporting user agents" > "in supporting user agents"

patrickhlauke: maybe even drop the "in supporting", because if the user agents were "supporting", they'd HAVE / expose those implicit semantics. so just "in user agents which lack..."

index.html
```diff
@@ -1176,7 +1201,8 @@ <h2 id="docconformance">
             </td>
             <td>
               <p>
-                If not provided an author defined accessible name by other methods: <strong class="nosupport">No `role`</strong>, and <strong>no `aria-*` attributes</strong> except
+                If not provided an author defined accessible name by other methods:
```
patrickhlauke: reads awkwardly. maybe "If no author defined accessible name is provided by other methods:"

index.html
```diff
@@ -2599,7 +2615,8 @@ <h2 id="docconformance">
                 <code>role=<a href="#index-aria-cell">cell</a></code> if a
                 descendant of a `table` element.
               </p>
-              <!--<p><code>role=<a href="#index-aria-cell">gridcell</a></code> if a descendant of a <code>table</code> elementwith <code>role=grid</code></p>-->
+              <!--<p><code>role=<a href="#index-aria-cell">gridcell</a></code> if a descendant
+                of a <code>table</code> elementwith <code>role=grid</code></p>-->
```
patrickhlauke:
```suggestion
                of a <code>table</code> element with <code>role=grid</code></p>-->
```

scottaohara: going to leave this as is.  going to be deleted / revised by another PR anyway.

scottaohara:
as far as i'm concerned, this PR will also close #228 and #223 per the new notes / examples added by this pr

scottaohara (approved):

stevefaulkner (approved):

PR #243, opened 2020-09-29
update allowances for image with empty alt

### LINKED ISSUE #229: use case for role=presentation/none and empty alt on an image
per this [html validator issue](https://github.com/validator/validator/issues/955) it got me wondering what use cases exist for declaring an image as presentational with `alt=""` but also needing to use an aria `role=none / presentation`.

As standard guidance is that authors should not need to use a role that is duplicative to the way in which the element is exposed, is this allowance needed?  Is this to mitigate against a browser / AT bug?  And if so, should that not be what's fixed instead of having a workaround?

@rianrietveld if you have a use case as to why both `alt=""` and `role=presentation` were used in your example in the validator thread, I'd be interested to hear it.

scottaohara:
commit to remove allowance for `role=presentation` and `role=none` on an ``.  Declaring these roles _should_ be considered redundant/unnecessary per the `alt=""`.  

If there are situations where an image with `alt=""` is not being set to decorative, (thinking an img with a .svg source in webkit) then authors can still use `aria-hidden=true` to force the element to be hidden to the a11y tree (and would have needed to use that anyway).

closes #229

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/243.html" title="Last updated on Oct 20, 2020, 1:21 PM UTC (7694d34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/243/8d10192...7694d34.html" title="Last updated on Oct 20, 2020, 1:21 PM UTC (7694d34)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1206,7 +1206,7 @@
             </td>
             <td>
               <strong class="nosupport">No `role` or `aria-*` attributes</strong> 
-              except `aria-hidden`.
+              except `aria-hidden="true"`.
             </td>
           </tr>
           <tr id="el-img-no-alt" tabindex="-1">
--- a/index.html
+++ b/index.html
@@ -1198,22 +1198,15 @@
           </tr>
           <tr id="el-img-empty-alt" tabindex="-1">
             <td>
-              <code><a>img</a></code> with <code><a data-cite=
-              "html/embedded-content.html#attr-img-alt">alt</a>=""</code>
+              <code><a>img</a></code> with 
+              <code><a data-cite="html/embedded-content.html#attr-img-alt">alt</a>=""</code>
             </td>
             <td>
               <a>No corresponding role</a>
             </td>
             <td>
-              <p>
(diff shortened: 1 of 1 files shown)


carmacleod:
Question: Did you want to allow `` for "img without alt attribute"?
(i.e. the example given in #229)

If so, might want to change the first sentence under "img without alt attribute" from:
> If no author-defined accessible name is provided by other methods: No role, and no aria-* attributes except aria-hidden.

to something like:
> If no author-defined accessible name is provided by other methods: Roles: none or presentation. No aria-* attributes except aria-hidden.

scottaohara:
@carmacleod no.  Then that could result in instances of the file name of the image being announced without the image role.

If an author can add `role=presentation` to their `` then it stands to reason they can add the shorter `alt=""` attribute and meet HTML's requirements to add an `alt`.

Issue #241, opened 2020-09-11
other aria-* attributes which should not be allowed on HTML elements


scottaohara:
for starters, aria-checked should not be used on a native `input type=checkbox/radio`  It's both unnecessary, and if out of alignment, the native HTML state will largely win (as it should) over the aria attribute (except in chrome where this rule is not implemented properly).

However, that's even more reason to make an explicit rule for this, so that developers are encouraged not to use the attribute unnecessarily and find themselves in a misalignment situation.

Similarly, should look at should not-ing `aria-selected` on `option`

mitchellevan:
I agree. Similarly, `aria-multiselect` should not be used on `<select multiple>`

scottaohara:
closed by #262

Issue #225, opened 2020-03-20
Incorrect or confusing description of aria-required usage


Haprog:
See this table: https://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html

Here's a screenshot with some unrelated rows deleted for easily showing the column headers with few related rows:
![image](https://user-images.githubusercontent.com/75155/77167724-39d85c80-6abf-11ea-809f-6c399b64ee66.png)

The highlighted piece of text in the third column seem to be contradicting since the second column defines that `aria-required="true"` should not be used on element with a `required` attribute.

Compare this text with the third column of the "disabled" and "readonly" rows which are more logical.

Should this instead say (similarly to disabled and readonly):
> Use the `required` attribute on any element that is [allowed the `required` attribute](https://html.spec.whatwg.org/multipage/input.html#attr-input-required) in HTML.
>
> Only use the `aria-required` attribute for elements that are **not allowed** to have a `required` attribute in HTML."

Or is the `aria-required` a more special case compared to `aria-disabled` or `aria-readonly`?

scottaohara:
closed by #262

PR #262, opened 2021-02-18
Separate html elements and attributes into different tables

### LINKED ISSUE #122: Consider to clarify aria-disabled usage
With regards to the `aria-disabled` attribute https://www.w3.org/TR/html-aria/#attr-disabled the first part is pretty clear: an element with a disabled attribute should not use `aria-disabled="true"` because its semantics is implicit and already conveyed by the `disabled` attribute.

However, I'm not fully sure I understand the second part:
> Only use the aria-disabled attribute for elements that are **not allowed** to have a disabled attribute in HTML5

which seems to completely prohibit to use `aria-disabled` on some elements, e.g. a button element.

For example, seems to me there are cases where using `aria-disabled` on a button _and keeping it focusable_ is beneficial. Avoiding a focus loss is one notable scenario. Discoverability of a feature is an important scenario as well.

The WAI-ARIA Authoring Practices 1.1 list some specific cases where making a control perceived as disabled but still focusable may be beneficial. A `disabled` attribute would make the control not focusable.

5.7 Focusability of disabled controls
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls	

Same in 1.2:
https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_disabled_controls

Also, in ARIA 1.1 it's stated that [aria-disabled](https://www.w3.org/TR/wai-aria-1.1/#aria-disabled) can be used in "All elements of the base markup". Same in [ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#aria-disabled).

I'm not an English native speaker so I might be a bit confused 🙂However, seems to me the current wording could be clarified a bit, as it sounds to completely prohibit the `aria-disabled` attribute for elements that are allowed to have a disabled attribute (e.g. buttons) while other specifications seems to allow it.

scottaohara:
Treat these tables separately so we can be more clear with guidance specific to roles vs attributes.

adds in new specific attribute call outs on elements such as checkbox, radio, options and selects which should not have specific aria-* attributes set.

revised/re-explained allowances for specific html attributes and their implicit aria attribute semantics are in the new second table.

This PR will close #122, #225 and #241

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/262.html" title="Last updated on Feb 21, 2021, 1:31 AM UTC (7370b3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/262/7d82cdd...7370b3f.html" title="Last updated on Feb 21, 2021, 1:31 AM UTC (7370b3f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,12 +66,15 @@
       <p>
         Web developers MAY use the ARIA `role` and `aria-*` attributes on
         <a data-cite="html/dom.html#elements">HTML elements</a>, in accordance
-        with the requirements described in [[wai-aria-1.1]], except where these
+        with the requirements described in [[wai-aria-1.1|WAI-ARIA]], except where these
         conflict with the
         <dfn><a data-cite="wai-aria-1.1#host_general_conflict">strong native semantics</a></dfn>
         or are equal to the
         <dfn><a data-cite="wai-aria-1.1#implicit_semantics">implicit ARIA semantics</a></dfn>
-        of a given HTML element. These constraints are intended to prevent
+        of a given HTML element. The <a>implicit ARIA semantics</a> for an HTML element are defined in the <dfn>[[html-aam-1.0|HTML Accessibility API Mappings]] specification.
+      </p>
+      <p>
+        The constraints defined in this specification are intended to prevent
         developers from making assistive technology products report nonsensical
         user interface (UI) information that does not represent the actual UI
         of the document.
@@ -79,7 +82,7 @@
       <p>
         Web developers MUST NOT use the ARIA `role` and `aria-*` attributes in
         a manner that conflicts with the semantics described in the
-        [[[#docconformance]]] table. It is NOT RECOMMENDED for web developers
+        [[[#docconformance]]] and [[[#docconformance-attr]]] tables. It is NOT RECOMMENDED for web developers
         to set the ARIA `role` and `aria-*` attributes to values that match the
         <a>implicit ARIA semantics</a> defined in the table. Doing so is unnecessary
         and can potentially lead to unintended consequences.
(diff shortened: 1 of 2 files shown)


scottaohara:
edit: scratching the note I had here. needs to be thought out more and i want to run some tests.

index.html
```diff
@@ -2315,9 +2335,10 @@ <h2 id="docconformance">
               <p>
                 Role: <a href="#index-aria-menu">`menu`</a>
               </p>
+              <p>Authors SHOULD NOT use the `aria-multiselect` attribute on a `select` element.</p>
```
mtrootyy: Incorrect: aria-multiselect
Correct: aria-multiselectable

index.html
```diff
@@ -2333,9 +2354,10 @@ <h2 id="docconformance">
               <p>
                 <strong class="nosupport">No `role`</strong>
               </p>
+              <p>Authors SHOULD NOT use the `aria-multiselect` attribute on a `select` element.</p>
```
mtrootyy: Incorrect: aria-multiselect
Correct: aria-multiselectable

index.html
```diff
@@ -1606,7 +1617,13 @@ <h2 id="docconformance">
                 <strong class="nosupport">No `role`</strong>
               </p>
               <p>
-                <a href="#index-aria-global">Global `aria-*` attributes</a> and
+                Authors SHOULD NOT use the `aria-haspopup` attribute on the indicated `input`s with a `list` attribute.
+              </p>
+              <p>
+                Authors MUST NOT use the `aria-haspopup` attribute with a value other than `list` on the indicated `input`s with a `list` attribute.
```
mtrootyy: Incorrect: with a value other than **`list`**
Correct: with a value other than **`listbox`**

index.html
```diff
@@ -2808,129 +2830,254 @@ <h2 id="docconformance">
...
+              </p>
+            </th>
+            <th>
+              Author guidance
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr id="att-checked" tabindex="-1">
+            <th>
```
patrickhlauke, line 2928: while semantically correct to use header cells for each row, the current standard styling makes this table look...rather unappealing. and it doesn't now match the structure of the other tables. might be worth just reverting to `<td>`s? or tweaking the styles and also modifying the other tables?

patrickhlauke (changes requested):

index.html
scottaohara: I made changes to a css file to modify the styling, but that isn’t being rendered as part of the preview

patrickhlauke: ah, gotcha. but yeah, then, changing all other tables to match? can probably be done with a semi-clever find/replace?

scottaohara: Yup. Was going to do that after this pr since that’d show “changes” everywhere then.

index.html
scottaohara: thank you @mtrootyy 

silly mistake on my part, was trying to get this wrapped up too quickly

patrickhlauke (changes requested):
Would still be good to make the change to the other tables to start rows with a `<th>` rather than just having that for the one new table?

patrickhlauke:
> Would still be good to make the change to the other tables to start rows with a `<th>` rather than just having that for the one new table?

happy to try and make a semi-clever find/replace for this if you want, @scottaohara

scottaohara:
@patrickhlauke i wouldn't mind but for the fact I have other PRs that may have unnecessary merge conflicts if we go doing that here.  I'd rather just get this and the other PRs merged in, and then this can easily be done.

patrickhlauke (approved):

patrickhlauke:
> @patrickhlauke i wouldn't mind but for the fact I have other PRs that may have unnecessary merge conflicts if we go doing that here. I'd rather just get this and the other PRs merged in, and then this can easily be done.

ok makes sense. i opened a separate issue to track this. so with that out of the way, i think this is good to go

stevefaulkner (approved):

PR #267, opened 2021-02-20
merge a and area rows

scottaohara:
merges the `a`/`area` and `a`/`area` w/out `href` rows into one row for each element.

we can do similar with `img` and `select` elements if we’re in agreement on this approach.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/267.html" title="Last updated on Feb 20, 2021, 10:32 PM UTC (ace8404)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/267/7d82cdd...ace8404.html" title="Last updated on Feb 20, 2021, 10:32 PM UTC (ace8404)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -186,12 +186,16 @@
         <tbody>
           <tr id="el-a" tabindex="-1">
             <td>
-              [^a^] with <a data-cite="html/links.html#attr-hyperlink-href">`href`</a>
+              [^a^]
             </td>
             <td>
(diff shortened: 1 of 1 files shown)


scottaohara:
actually going to close this one for now. it's not necessary to do and might actually make some other work easier if we don't merge them.  going to have a think on that.

going to not delete the branch for now though, just in case.

PR #285, opened 2021-03-08
chore: remove more data-cite= usage

marcoscaceres:
Drops more use of `data-cite=` ... ideally, there should not be any data-cite= used, specially for HTML things. 

Fixed up a few more linking issues to HTML spec. 

Note: this is blocked on https://github.com/whatwg/html/pull/6464

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/285.html" title="Last updated on Mar 13, 2021, 5:49 PM UTC (9f0030c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/285/72cc9b4...9f0030c.html" title="Last updated on Mar 13, 2021, 5:49 PM UTC (9f0030c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2868,7 +2868,7 @@
         represented in roles, states and properties not provided by ARIA, and
         exposed to users of assistive technology via accessibility APIs. It is
         therefore recommended that developers add a `role` attribute to a
-        semantically neutral element such as a `div` or `span`, rather than
+        semantically neutral element such as a [^div^] or [^span^], rather than
         overriding the semantics of the listed elements.
       </p>
       <div class="note">
@@ -2881,8 +2881,8 @@
           add accessibility information to HTML elements using the Accessible
           Rich Internet Applications specification (ARIA 1.1).
           </li>
-          <li>[[[wai-aria-practices-1.2]]] - An author's guide to understanding
-          and implementing Accessible Rich Internet Applications.
+          <li>[[[wai-aria-practices-1.2]]] - A developer's guide to
+            understanding and implementing Accessible Rich Internet Applications.
           </li>
         </ul>
       </div>
@@ -2920,7 +2920,7 @@
         Document conformance requirements for use of ARIA attributes with HTML attributes
       </h3>
       <p>
-        Unless otherwise stated, developers MAY use `aria-*` attributes in place of their HTML equivalents on HTML elements where the `aria-*` semantics would be expected. For example, developers MAY use `aria-disabled=true` on a `button` element, rather than the `disabled` attribute. However, developers SHOULD NOT use both the native HTML attribute and the `aria-*` attribute together, and MUST NOT use the these features together when their values are in opposition to each other. As stated in
+        Unless otherwise stated, developers MAY use `aria-*` attributes in place of their HTML equivalents on HTML elements where the `aria-*` semantics would be expected. For example, developers MAY use `aria-disabled=true` on a [^button^] element, rather than the `disabled` attribute. However, developers SHOULD NOT use both the native HTML attribute and the `aria-*` attribute together, and MUST NOT use the these features together when their values are in opposition to each other. As stated in
         <a data-cite="wai-aria-1.1#host_general_conflict">WAI-ARIA's Conflicts with Host Language Semantics</a>,
         user agents MUST ignore WAI-ARIA attributes and use the host language
         (HTML) attribute with the same <a>implicit ARIA semantics</a>.
@@ -3045,7 +3045,7 @@
           </tr>
           <tr id="att-max" tabindex="-1">
             <th>
-              Any element where the `max` attribute is allowed: `meter` [^meter/max^], `progress` [^progress/max^], and `input` [^input/max^]. 
+              Any element where the `max` attribute is allowed: `meter` [^meter/max^], `progress` [^progress/max^], and `input` [^input/max^].
             </th>
             <td>
               `aria-valuemax="..."`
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Travis will keep failing until the terms are exported from HTML. I'll keep an eye on it in case it falls out sync with the base branch (no stress if it does... easy to fix).

marcoscaceres:
@scottaohara, HTML is not exporting what we need, so you want to take this one as is? (as long as I didn't screw any of them up.... let me know if I did). Having another look based on the discussion in the other bug.

index.html
```diff
@@ -65,7 +65,7 @@ <h2 id="rules-wd">
       </h2>
       <p>
         Developers MAY use the ARIA `role` and `aria-*` attributes on
-        <a data-cite="html/dom.html#elements">HTML elements</a>, in accordance
+        [=HTML elements=], in accordance
```
scottaohara: this update changes the link destination, so this needs to be reverted or updated 
previous: https://html.spec.whatwg.org/multipage/dom.html#elements
new: https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements

index.html
```diff
@@ -175,7 +175,7 @@ <h2 id="docconformance">
         The following table provides normative per-element document-conformance
         requirements for the use of ARIA markup in HTML documents and describes
         the <a>implicit ARIA semantics</a> that apply to
-        <a data-cite="html/dom.html#elements">HTML elements</a> as defined in [[html-aam-1.0|HTML AAM]].
+        [=HTML elements=] as defined in [[html-aam-1.0|HTML AAM]].
```
scottaohara, line 178: similar with previous comment.  not where i'd want this HTML elements link to go

index.html
```diff
@@ -1245,7 +1243,7 @@ <h2 id="docconformance">
           </tr>
           <tr id="el-img-no-alt" tabindex="-1">
             <th>
-              [^img^] without <code><a data-cite="html/images.html#unknown-images">alt</a></code> attribute
+              [^img^] without [^img/alt^] attribute
```
scottaohara: this changes the link.
now goes to: https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt
but should go to https://html.spec.whatwg.org/multipage/images.html#unknown-images

this needs to be fixed

index.html
```diff
@@ -3096,8 +3089,8 @@ <h3 id="docconformance-attr">
           </tr>
           <tr id="att-readonly" tabindex="-1">
             <th>
-              Any element where the <a data-cite="html/input.html#attr-input-readonly">`readonly`</a>
-              attribute is allowed
+              Any element where the `readonly`
+              attribute is allowed: [^face/readonly^],
```
scottaohara: this changes the link
was: https://html.spec.whatwg.org/multipage/input.html#attr-input-readonly
now: https://html.spec.whatwg.org/multipage/custom-elements.html#attr-face-readonly

could link to both, but more importantly should be linking to the previous link as that's what is more applicable here

scottaohara (changes requested):
checked everything.  a few instances of unexpected changing of link destinations that need updating

index.html
marcoscaceres: I think the new link is correct tho. It's the HTML elements, including those in the XHTML namespace. The old link is meaningless, at it points to a section, not to a definition?

index.html
marcoscaceres: As above. I also think we should fix the wording of this to:

> The following table provides normative per-element document-conformance  requirements for the use of ARIA markup in HTML documents and describes the <a>implicit ARIA semantics</a>, as defined in [[html-aam-1.0|HTML AAM]], that apply to [=HTML elements=] . 

Otherwise it reads like HTML AAM defines the HTML elements, which is not right.

WDYT?

index.html
marcoscaceres: The #unknown-images link is not exported, so we can't link to it unfortunately. It would also be inconsistent to link to  link to [^img/alt^] in some places but link to somewhere else to others. 

Let me see if I can find something else to link to, but I think [^img/alt^] is also correct here.

index.html
scottaohara: understand your point, but the link is _not_ meaningless in context of this spec and the information that's presented there.  

would suggest the following then

"Developers MAY use the ARIA `role` and `aria-*` attributes to change the exposed meaning (semantics) of [=HTML elements=], in accordance...

where in the above "semantics" would link to https://html.spec.whatwg.org/multipage/dom.html#semantics-2 (which is really where the previous link should have gone i suppose).

so we get that new link in here, then we can move this along.

index.html
scottaohara: well html aam defines the mappings, which is right.  but regardless of that, per my comment on the previous instance of this, we can make this change if we get the semantics link in place.

index.html
scottaohara: agree with your point on having `alt` alone not link to different places.  so i would suggest the link text become "without `alt`" so that it can continue to link to that content.

index.html
scottaohara: made a commit to resolve this per my concerns

index.html
scottaohara: updated link text so as to keep original link

index.html
scottaohara: revised the wording of this cell so both could be linked to

scottaohara (approved):

PR #299, opened 2021-03-24
Add treeitem and gridcell to list of allowed button roles

### LINKED ISSUE #298: Allow gridcell and treeitem roles on button elements
In the [Rules of ARIA attribute usage by HTML element ](https://w3c.github.io/html-aria/#docconformance), `<button>` already allows every child role of [composite widgets that manage focus](https://w3c.github.io/aria/#managingfocus_authors) except for tree and grid.

There are benefits in using `<button>` as a base of interaction in these widgets, such as:
- Enter/space triggers the click event
- Certain colors and interaction styles are applied automatically in forced color modes like Windows High Contrast Mode.

Those benefits also apply to `gridcell`s and `treeitem`s, so it seems like it would be valuable to add those as well.

Happy to do a PR for this :)

smhigley:
Resolves #298 

Adds `gridcell` and `treeitem` to the list of allowed button roles, so that it can be used as a base within all compound focus-managing widget roles.

***
<a href="https://pr-preview.s3.amazonaws.com/smhigley/html-aria/pull/299.html" title="Last updated on Jul 23, 2021, 12:41 PM UTC (ad9ddbc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/299/e8b51bc...smhigley:ad9ddbc.html" title="Last updated on Jul 23, 2021, 12:41 PM UTC (ad9ddbc)">Diff</a>


scottaohara:
hey @smhigley.

Right now we're holding on accepting PRs as we finish the implementation report / go through wide review.  But I think this is definitely worth looking at.  thank you

scottaohara:
unfortunately after giving this some more thought / talking with Sarah about this, it turns out updating to allow these roles wouldn't really work out.  For instance, both a `gridcell` and `treeitem` would allow for nested elements – a `treeitem` in particular could have an entire nested tree within, where as a `gridcell` could have nested links, buttons, etc..  But HTML's `button` element does not allow for the nesting of interactive elements, and the native behavior of the `<button>` would likely cause issues if interactive elements were nested.

PR #312, opened 2021-04-29
Editorial changes to tr, th, td for consistency with other roles

### LINKED ISSUE #311: Editorial changes to tr, th, td for consistency with other roles
Given the "slightly softer" stance around unnecessary explicit roles in https://github.com/w3c/html-aria/issues/236, the entries for `tr`, `th`, and `td` look a bit strange now, because they say:
> No role if the ancestor table element has role=table, grid, or treegrid; otherwise any role.

when they could more simply say "Any role".

As far as I can see, these are the only roles that explicitly say that you're not allowed to use an explicit role. :)

Also, there's a bit of inconsistency in the ordering of the wording for the implied role for `th`; and the implied role for `tr` doesn't say anything about the ancestor table element.

carmacleod:
- fixes #311

***
<a href="https://pr-preview.s3.amazonaws.com/carmacleod/html-aria/pull/312.html" title="Last updated on Apr 29, 2021, 8:38 PM UTC (68e88df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/312/2f7d674...carmacleod:68e88df.html" title="Last updated on Apr 29, 2021, 8:38 PM UTC (68e88df)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2623,9 +2623,7 @@
             </td>
             <td>
               <p>
-                <strong class="nosupport">No `role`</strong> if the ancestor `table`
-                element has `role=table`, `grid`, or `treegrid`; otherwise
-                <a><strong>any</strong> `role`</a>.
+                <a><strong>Any</strong> `role`</a>
               </p>
               <p>
                 <a href="#index-aria-global">Global `aria-*` attributes</a> and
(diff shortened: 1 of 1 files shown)


carmacleod:
Actually, never mind this - bogus thinking on my part.
I was thinking that the wording was just to disallow using explicit role=row, etc. But of course tr, th, td can't have any other role when their table ancestor is a table, grid or treegrid.
Sorry for the swirl.

PR #321, opened 2021-05-13
Remove aria-invalid from the ARIA attributes in place of HTML equivalents table

patrickhlauke:
Discussed with Scott today, and we agreed:

* there is no equivalent HTML attribute that serves the same purpose - it touches on HTML element DOM *properties* (so it doesn't belong in the table that compared ARIA attributes vs equivalent HTML attributes in the first place)
* testing this would require validators to do dynamic checking of the DOM and the element *properties*. while a contrived static test case can be made, it would only be an artificial test

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/321.html" title="Last updated on May 13, 2021, 5:30 PM UTC (31a7310)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/321/04a33b7...31a7310.html" title="Last updated on May 13, 2021, 5:30 PM UTC (31a7310)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3144,26 +3144,6 @@
                 </p>
               </td>
             </tr>
-
-
-            <tr id="IDL-ValidityState" tabindex="-1">
-              <th>
-                Element that is a <a data-cite=
-                "html/form-control-infrastructure.html#candidate-for-constraint-validation">candidate for
-                constraint validation</a> but that does not <a data-cite=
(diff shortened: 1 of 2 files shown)


scottaohara (approved):

scottaohara:
ayup. if we can think up a way to get something like this back in for the next version, that'd be good. but can't have it in for now.

PR #340, opened 2021-07-27
Revise leading paragraph to attribute conformance section

### LINKED ISSUE #339: Unclear interaction of "MUST NOT use opposing values" requirement with aria-checked and aria-hidden
## Potential  Issue

In "3.1 Requirements for use of ARIA attributes in place of equivalent HTML attributes" the leading paragraph says:

> However, authors SHOULD NOT use both the native HTML attribute and the aria-* attribute together, and MUST NOT use the these features together when their values are in opposition to each other
https://www.w3.org/TR/html-aria/#docconformance-attr

This is an absolute prohibition on using opposing values per RFC 2119.

The 'Rules of ARIA attribute usage by HTML feature' table below this restates the "MUST NOT use these ... in opposition"  requirement for aria-disabled, aria-readonly, aria-required and strengthens it for aria-valuemax, aria-valuemin

However the intent seems to be the "MUST NOT use these ... in opposition" is relaxed for aria-checked and aria-hidden, but the wording isn't explicit about this:

- `aria-hidden=false` and `hidden` together is an open issue (https://github.com/w3c/html-aria/issues/270), but my reading of the current wording is the "MUST NOT use the these features together when their values are in opposition to each other" in the leading paragraph applies since there's no wording indicating an exception for `aria-hidden` 

- I think the intent is the "SHOULD NOT use the aria-checked attribute on any element where the checkedness of the element can be in opposition to the current value of the aria-checked attribute" overrides the  "MUST NOT use these ... in opposition" in the leading paragraph, but this is unclear since the MUST NOT is an absolute prohibition. The tests in https://w3c.github.io/html-aria/tests/checked-test.html flag a SHOULD NOT error for `aria-checked=true` without `checked`.

## Possible Solutions

1) One way of resolving this would be adding "except as noted below"  or similar to the "MUST NOT use the these features together when their values are in opposition to each other" in the leading paragraph, then add some explicit wording where the requirement is relaxed.

2) Alternatively, it might be better to remove the "MUST NOT use these ... in opposition" in the leading paragraph since the requirements seem very specific to each attribute - and https://github.com/w3c/html-aria/issues/338 may make these even more specific. If there's still a need for a generic "MUST NOT use opposing values" fallback that could be added as an "otherwise" row to the end of the table.

**Edit:** The second option is probably better since the requirements for each attribute are in one place (which makes implementation easier). Keeping stuff closer together also works better with Github diffs - when reviewing PRs it's easy to miss important lines that should have been changed because they're too far away from the changed lines to appear in the diff.

scottaohara:
Revises leading paragraph to remove “must not” wording, but indicate that additional requirements can be made on a per attribute basis (where the must not rules remain).

closes #339

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/340.html" title="Last updated on Jul 27, 2021, 8:01 PM UTC (f2b79d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/340/e8b51bc...f2b79d2.html" title="Last updated on Jul 27, 2021, 8:01 PM UTC (f2b79d2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2859,7 +2859,7 @@
           As stated in
           <a data-cite="wai-aria-1.1#host_general_conflict">WAI-ARIA's Conflicts with Host Language Semantics</a>,
           user agents MUST ignore WAI-ARIA attributes and use the host language
-          (HTML) attribute with the same <a>implicit ARIA semantics</a>. For this reason, authors SHOULD NOT specify both the native HTML attribute and the equivalent `aria-*` attribute on an element. Please review each attribute for any further author requirements specific to each attribute.
+          (HTML) attribute with the same <a>implicit ARIA semantics</a>. For this reason, authors SHOULD NOT specify both the native HTML attribute and the equivalent `aria-*` attribute on an element. Please review each attribute for any further author specific requirements.
         </p>
         <p>
           The following table represents HTML elements and their attributes which have `aria-*` attribute parity.
--- a/index.html
+++ b/index.html
@@ -2858,8 +2858,7 @@
         <p>
           As stated in
           <a data-cite="wai-aria-1.1#host_general_conflict">WAI-ARIA's Conflicts with Host Language Semantics</a>,
-          user agents MUST ignore WAI-ARIA attributes and use the host language
-          (HTML) attribute with the same <a>implicit ARIA semantics</a>. For this reason, authors SHOULD NOT specify both the native HTML attribute and the equivalent `aria-*` attribute on an element. Please review each attribute for any further author specific requirements.
+          when HTML elements use <em>both</em> `aria-*` attributes and their host language (HTML) equivalents, user agents MUST ignore the WAI-ARIA attributes – the native HTML attributes with the same <a>implicit ARIA semantics</a> take precedence. For this reason, authors SHOULD NOT specify both the native HTML attribute and the equivalent `aria-*` attribute on an element. Please review each attribute for any further author specific requirements.
         </p>
         <p>
           The following table represents HTML elements and their attributes which have `aria-*` attribute parity.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2853,10 +2853,13 @@ <h3 id="docconformance-attr">
           Requirements for use of ARIA attributes in place of equivalent HTML attributes
         </h3>
         <p>
-          Unless otherwise stated, authors MAY use `aria-*` attributes in place of their HTML equivalents on HTML elements where the `aria-*` semantics would be expected. For example, authors MAY use `aria-disabled=true` on a [^button^] element, rather than the `disabled` attribute. However, authors SHOULD NOT use both the native HTML attribute and the `aria-*` attribute together, and MUST NOT use the these features together when their values are in opposition to each other. As stated in
+          Unless otherwise stated, authors MAY use `aria-*` attributes in place of their HTML equivalents on HTML elements where the `aria-*` semantics would be expected. For example, authors MAY specify `aria-disabled=true` on a [^button^] element, while also implementing the necessary scripting to functionally disable the `button`, rather than the use `disabled` attribute.
+        </p>
+        <p>
+          As stated in
```
patrickhlauke, line 2859: i'd turn this around slightly, along the lines of

```
As stated in ..., when HTML elements use <em>both</html> `aria-*` attributes and their HTML equivalents, user agents MUST ignore the WAI-ARIA attributes - the native HTML attributes take precedence. For this reason, ...
```

patrickhlauke (changes requested):

index.html
scottaohara: @patrickhlauke updated.

patrickhlauke (approved):

PR #353, opened 2021-09-23
Correction: revise wbr and br element allowances

### LINKED ISSUE #332: wbr element
presently allows any role.  This seems incorrect and rather should not allow any role.

Additionally, should only allow for `aria-hidden` attribute, as other global attributes don't make sense for this element.

A similar global attribute update needs to be made for the `br` element.

### LINKED ISSUE #346: Allow aria-hidden=true on picture element
presently the picture element allows for no roles or aria-* attributes, which generally makes sense because the picture element itself is not an accessible object.   That said, there could be a situation where a picture that contains a decorative image needs to be hidden..... and for whatever reason the expected `alt=""` on the fallback image cannot be used, but `aria-hidden=true` on the picture element could.   Upon testing this just now, seems like this would be a viable option as Chromium browsers, Firefox and Webkit all ignore the child `` in this situation.

scottaohara:
closes #332

Revise the allowances for the `br` element - only allow `aria-hidden` attribute, rather than all global attributes.

Revise allowances for `wbr` to match `br`.  Neither of these elements allow for styling from authors, nor do they allow for child content (or content from CSS - again no styling).  So allowing for any roles/attributes beyond presentation or for hiding would not be a wise idea.

[html validator](https://github.com/validator/validator/issues/1241) 
- [x] role update
- [x] allowed attributes update

[ibm equal access accessibility checker](https://github.com/IBMa/equal-access/issues/507)
- [x] role update
- [x] allowed attributes update

[axe-core](https://github.com/dequelabs/axe-core/issues/3177)
- [x] role update
- [ ] allowed attributes update

[arc toolkit](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/issues/42)
- [x] role update
- [x] allowed attributes update

html validator and axe issues are still open in regards to limiting allowed `aria-*` attributes on these elements to `aria-hidden` alone.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/353.html" title="Last updated on Dec 8, 2021, 1:04 AM UTC (e1dd099)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/353/3ef70c1...e1dd099.html" title="Last updated on Dec 8, 2021, 1:04 AM UTC (e1dd099)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -539,7 +539,7 @@
                 <a href="#index-aria-presentation">`presentation`</a>
                 or <a href="#index-aria-none">`none`</a>.
               </p>
-              <p>
+              <p class="proposed addition">
                 Authors MAY specify the <a data-cite="wai-aria-1.1#aria-hidden">`aria-hidden`</a> attribute on the `br` element.
                 Otherwise, no other allowed `aria-*` attributes.
               </p>
@@ -2788,15 +2788,17 @@
               <a>No corresponding role</a>
             </td>
             <td>
-              <p>
-                Roles:
-                <a href="#index-aria-presentation">`presentation`</a>
-                or <a href="#index-aria-none">`none`</a>.
-              </p>
-              <p>
-                Authors MAY specify the <a data-cite="wai-aria-1.1#aria-hidden">`aria-hidden`</a> attribute on the `wbr` element.
-                Otherwise, no other allowed `aria-*` attributes.
-              </p>
+              <div class="proposed correction">
+                <p>
+                  Roles:
+                  <a href="#index-aria-presentation">`presentation`</a>
+                  or <a href="#index-aria-none">`none`</a>.
+                </p>
+                <p>
+                  Authors MAY specify the <a data-cite="wai-aria-1.1#aria-hidden">`aria-hidden`</a> attribute on the `wbr` element.
+                  Otherwise, no other allowed `aria-*` attributes.
+                </p>
+              </div>
             </td>
           </tr>
         </tbody>
--- a/index.html
+++ b/index.html
@@ -540,8 +540,8 @@
                 or <a href="#index-aria-none">`none`</a>.
               </p>
               <p>
-                <a data-cite="wai-aria-1.1#global_states">Global `aria-*` attributes</a> and
-                any `aria-*` attributes applicable to the allowed roles.
+                Authors MAY specify the <a data-cite="wai-aria-1.1#aria-hidden">`aria-hidden`</a> attribute on the `br` element.
+                Otherwise, no other allowed `aria-*` attributes.
               </p>
             </td>
           </tr>
@@ -2783,11 +2783,13 @@
             </td>
             <td>
               <p>
-                <a><strong>Any</strong> `role`</a>
+                Roles:
+                <a href="#index-aria-presentation">`presentation`</a>
+                or <a href="#index-aria-none">`none`</a>.
               </p>
               <p>
-                <a data-cite="wai-aria-1.1#global_states">Global `aria-*` attributes</a> and
-                any `aria-*` attributes applicable to the allowed roles.
+                Authors MAY specify the <a data-cite="wai-aria-1.1#aria-hidden">`aria-hidden`</a> attribute on the `wbr` element.
+                Otherwise, no other allowed `aria-*` attributes.
               </p>
             </td>
           </tr>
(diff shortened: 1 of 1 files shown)


scottaohara:
This can now be merged after the 1.0 spec officially meets Rec status.

index.html
```diff
@@ -540,8 +540,8 @@ <h2 id="docconformance">
                 or <a href="#index-aria-none">`none`</a>.
               </p>
               <p>
```
marcoscaceres: If this a correction, wrap with `<div "correction.proposed">`.

index.html
```diff
@@ -2783,11 +2783,13 @@ <h2 id="docconformance">
             </td>
             <td>
               <p>
-                <a><strong>Any</strong> `role`</a>
+                Roles:
```
marcoscaceres: As above... "addition.proposed" or "correction.proposed"?

index.html
scottaohara: edit: nm on my previous question.  talked to Leonie and understand that these need to be marked up as `<div class="class class">`

index.html
scottaohara: this one is actually both.  a correction (roles) AND addition (attribute allowances)

if, per my above question about adding the classes, this would be awkward visual styling to treat these separately, having two highlighted boxes right next to each other.

this approach also seems like it'll make other updates, where maybe a single role has been added to a paragraph, require the entire paragraph be contained within one of these divs.  that won't make it very easy to identify what specifically was updated... but sure will make a lot of big colorful boxes around the spec for the various things we have planned.

marcoscaceres: We should add it to a span if needed. We can then update base.css to support that. The W3C Process is not here to cause us grief or get in the way - we can adapt as needed. 

Cc @plehegar

plehegar: The process requires us to "identify the specific candidate amendments under review as proposed amendments (proposed corrections/proposed additions)." Pubrules has a [rule](https://github.com/w3c/specberus/blob/main/lib/rules/sotd/rec-addition.js#L43) to check the wording in the SOTD but doesn't go beyond that.

I'm thinking this is a corner case. My take would be to use the CSS style for the addition, ie .addition.proposed since it seems to be more relevant from an IP perspective. You can always add a note to it if you need to highlight the subtle aspects to the reviewer. (cc @frivoal).

scottaohara: i'm not sure... here are visuals I created, using dev tools to add in the classes, based on what i'm hearing as the suggestions (using PR #367 as an example)...

per what @marcoscaceres was saying, we might need to do stuff like this for various sections where we are adding allowed roles to a particular element. a single word in a paragraph of roles.
highlighting that group has been added to allowed roles
This would not be a corner case for this spec.  

Per what I'm hearing from @plehegar, the expectation would instead be to highlight the entire paragraph, and then add a note (using placeholder text in this example) to indicate the single addition of the group role?
visually highlight entire role allowances paragraph. adds a separate note to indicate the 'group' role was added
This seems heavy handed, and the spec would get full of various colorful boxes, quickly, with notes to describe single word changes.

Would we also need to follow the second example, for instance, if we were to remove a role from being allowed?

e.g., instead of adding a role to the above, say "feed" was removed.  Highlight the paragraph and then add a note saying "feed role has been removed?"

We will need to do something like this as ARIA 1.2 is deprecating the `directory` role, and this spec will update to remove it as an allowed role for some elements per the role's deprecation.

I understand the need for indicating what's been changed in the spec, but I'm really concerned with how distracting and messy this will make this spec per the above.   Could we potentially add in these classes, but suppress/modify the styling, and make sure our change log linked specifically to the necessary table rows where content has been updated, per these example?

I am far less concerned with using these styles to indicate changes in other sections of the spec (not in small table cells), where these particular styles seem to be better suited for.

index.html
marcoscaceres: yes, correct.. `"class class"` is right. Should try things before suggesting them 🤪

index.html
marcoscaceres: Agree that having to highlight and add a note seems like a lot of work - if there are a lot of these little changes.  

I'd personally go with the `span` option, which at least allows people to see what's changed. We might still want to give a summary of what's changed, but that can be in the change log (which can even be added directly in to the SoTD). 

For larger features, they can be

scottaohara:
As 3 of the 4 checkers have already updated the role allowances for `wbr`, and intent to implement the additional `aria-*` allowances for `br` and `wbr` has met no opposition, I'm going to merge this into the spec.

Issue #325, opened 2021-05-29
Consider allowing roles on <area> with no href attribute


scottaohara:
Creating this issue to replace #317, as there is no need to resolve that issue for 1.0 specification, but we should consider this suggestion going forward.

>`<a>` with no `href` takes any role, but `<area>` with no `href` takes no roles. These are similar and seem like they should be consistent. `<area role=button>` seems like it would be useful, as would checkbox and other widget roles.

scottaohara:
Spent quite a bit of time reviewing this today.  Due to the `area` element's inability to be styled effectively, that it cannot contain any child elements (save for css pseudo-element content - but even that doesn't render near the actual `area`, but rather adjacent to the image map)... i really don't see how this would be a good change.  though the concept of `a` and `area` without `href` are similar, the actual rendered result is quite different.

That said I *can* see a potential use case of allowing `role=button` or `role=link` (ideally with `tabindex=0`, but there can also be legitimate use cases for needing to expose these roles but not wanting them in the tab order - per other elements that perform the same functionality).   

cc @stevefaulkner  and @patrickhlauke, what say you two on this?

patrickhlauke:
discussed at meeting today with @scottaohara and @SteveFaulkner - agreed to allow `role="button"` and `role="link"` (with `tabindex="0"`) but not other roles due to weird browser behaviour/treatment of `<area>`

scottaohara:
specifically due to the fact that no styling is allowed for these elements, or allow for nested content in away that would make sense for any other role here.

PR #360, opened 2021-10-05
Addition: update allowances for area no href

scottaohara:
closes issue #325

This update allows authors to specify the `link` or `button` roles on an `area` with no `href` attribute.  

Any more complex widget, particularly any that would need to convey a visual state or updated styling, would be far more practical for authors to create another accessible control instead.  As these elements do not allow nested content (and even CSS pseudo element content appears in a different location than where the `area'`s coordinates place it, no container-type role would make sense here either.

Ideally, authors would be using a `tabindex` on these elements when specifying a role.... but as there could be situations where a role needs to be conveyed, but it is not necessary to make the particular element keyboard accessible (alternate mechanism available), the requirement for including a `tabindex` was left off this PR.  Additionally, we make no such requirements for other widget roles, so on second thought it seemed out of place to require one for this role.

---

- [x] [html validator](https://github.com/validator/validator/issues/1249)
- [x] [ibm equal access accessibility checker](https://github.com/IBMa/equal-access/issues/515)
- [x] [axe-core](https://github.com/dequelabs/axe-core/issues/3266)
- [x] [arc toolkit (PR filed)](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/pull/43)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/360.html" title="Last updated on Nov 16, 2021, 1:25 PM UTC (ffb41af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/360/0c12795...ffb41af.html" title="Last updated on Nov 16, 2021, 1:25 PM UTC (ffb41af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -337,9 +337,14 @@
             </th>
             <td><a>No corresponding role</a></td>
             <td>
-              <p><strong class="nosupport">No `role`</strong></p>
-              <p><a data-cite="wai-aria-1.1#global_states">Global `aria-*` attributes</a>
-              and any `aria-*` attributes applicable to the allowed roles.</p>
+              <p>Roles:
+                <a href="#index-aria-button">`button`</a>
+                or <a href="#index-aria-link">`link`</a>
+              </p>
+              <p>
+                <a data-cite="wai-aria-1.1#global_states">Global `aria-*` attributes</a> and
+                any `aria-*` attributes applicable to the allowed roles.
+              </p>
             </td>
           </tr>
           <tr id="el-article" tabindex="-1">
--- a/index.html
+++ b/index.html
@@ -4176,7 +4176,7 @@
         A conformance checker MAY define their own terminology, and level or levels of
         severity, when surfacing document failures to conform to this specification.
       </p>
-</section>
+    </section>
     <section id="priv-sec" class="informative">
       <h2>
         Privacy and security considerations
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

index.html
```diff
@@ -337,9 +337,14 @@ <h2 id="docconformance">
             </th>
             <td><a>No corresponding role</a></td>
             <td>
-              <p><strong class="nosupport">No `role`</strong></p>
-              <p><a data-cite="wai-aria-1.1#global_states">Global `aria-*` attributes</a>
-              and any `aria-*` attributes applicable to the allowed roles.</p>
+              <p>Roles:
```
marcoscaceres: Correction or addition?

marcoscaceres:
Just had a thought... maybe we can prefix the commit messages with "correction:" and "addition:"? That could also help us identify what's what.

scottaohara:
@marcoscaceres is every edit one of those?  from now on and forever?  most of these, i suppose would additions i suppose.  corrections would indicate that we should have had the spec perfect from the onset?

marcoscaceres:
> @marcoscaceres is every edit one of those? from now on and forever? 

Not every edit... some are a "chore" (e.g., a typo, grammar, or markup fix). However, most would be either an "addition" (new feature) or "correction" (bug fix). 

> most of these, i suppose would additions i suppose. corrections would indicate that we should have had the spec perfect from the onset?

"correction" means, "we found and fixed a bug". There is never an assumption that a spec is perfect: on the contrary, the assumption is that specs have bugs, and "corrections" fixes those bugs.

PR #372, opened 2021-11-10
Correction: update aria-checked allowances

### LINKED ISSUE #361: Update allowances for aria-checked on inputs
Presently we say that authors SHOULD NOT use aria-checked on radios and checkboxes.

After discussing https://github.com/w3c/aria/issues/1622 in today's ARIA WG meeting, and the problems that using `aria-checked` on native checkboxes can have - particularly due to `aria-che3cked=mixed` and bugs that exist now with Chromium browsers where `input type=checkbox checked aria-checked=false` will expose 'false' even though the control is actually 'checked', this rule needs to be updated to MUST NOT.

scottaohara:
closes  #361

Per the peculiar nature of `aria-checked`, where it allows for three different states on checkboxes (false, true, mixed), but native HTML only allows for true/false, and then has a separate IDL attribute to handle 'mixed', per ARIA's normative rules about host language attributes having higher importance, authors really must not use the `aria-checked` attribute on native HTML elements that allow it.  **Particularly** `input type=checkbox` which, if following the requirements of ARIA, as the checked/unchecked MUST always win out, and thus `aria-checked` MUST always be ignored, even if set to `aria-checked=mixed`.

These rule update are part of what will resolve the ARIA issue I opened - https://github.com/w3c/aria/issues/1622

---

- [x] [html validator](https://github.com/validator/validator/issues/1267)
- [x] [ibm equal access accessibility checker](https://github.com/IBMa/equal-access/issues/347)
- [x] [axe-core](https://github.com/dequelabs/axe-core/issues/3307)
- [x] [arc toolkit](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/pull/47)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/372.html" title="Last updated on Dec 8, 2021, 12:48 AM UTC (196abd0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/372/3ef70c1...196abd0.html" title="Last updated on Dec 8, 2021, 12:48 AM UTC (196abd0)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/372.html" title="Last updated on Apr 16, 2022, 3:08 PM UTC (ade71e7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/372/f1cae02...ade71e7.html" title="Last updated on Apr 16, 2022, 3:08 PM UTC (ade71e7)">Diff</a>


dylanb:
This change will make it extremely difficult for implementors to implement checkboxes that are styled AND are accessible. Currently most component libraries that implement accessible styled checkboxes use a real checkbox elements for the focus but overlay a visibly styled element of some sort.

I think that this is a change that takes the web platform in the wrong direction.

JAWS-test:
> Currently most component libraries that implement accessible styled checkboxes use a real checkbox elements for the focus but overlay a visibly styled element of some sort.

That is true

> This change will make it extremely difficult for implementors to implement checkboxes that are styled AND are accessible

This is not true as the frameworks do not require ARIA. There is the invisible HTML checkbox (without ARIA) and the visible checkbox (which is marked with aria-hidden=true at most, depending on the implementation)

scottaohara:
This update is being made to better reflect reality, and to inform authors that what they are doing now is not consistently accessible. Using `aria-checked` on native checkboxes and radio buttons is not respected in Firefox and Safari, which includes the `aria-checked=mixed` state. These browsers are correctly following the ARIA spec, as it has been written since version 1.0.  Native attributes win out over `aria-*` equivalents.

Authors using `aria-checked=mixed` on a native checkbox are presently only exposing this in Chromium browsers, due to their incorrect implementation of ARIA.  

Authors should be using the `indeterminate` IDL attribute to set their checkboxes to the mixed state - which per standard visually hidden checkbox techniques, [would work similarly](https://codepen.io/scottohara/pen/JjyVQJx) to what authors are doing with their `[aria-checked=mixed] + span` selectors now.


Further discussion on whether `aria-checked` needs to be a special case (because it actually is compared to other `aria-*` attributes, since it combines the checkboxes native checked and indeterminate attributes into one `aria-*` attribute) can occur in w3c/aria#1622.

stevefaulkner (approved):

scottaohara:
verified these rules are included in the IBM Equal Access preview rules.  This now meets the 2 implementations needed to be merged into the spec.

Issue #358, opened 2021-10-03
Clarification: update definition of 'any role'


scottaohara:
The current definition of 'any role' states:
> ...Any role it indicates that any role value apart from the implicit ARIA semantics role value, MAY be used.

this definition should be updated to indicate that this is not a carte blanche invitation to truly add 'any role' to an element.

For instance, we should make mention of the fact that any specified role must also follow the rules of use as outlined in the ARIA specification.  e.g., `<i role=gridcell>` *is  valid* per `<i>` allowing 'any role', BUT it would be invalid to add this role to an`<i>` if that element were not owned by a `role=row`, which that would then need to be owned by a `role=grid`.

scottaohara:
note that #391 has added some updates to the paragraph that initially describes 'any role'.

PR #333, opened 2021-06-27
revise casing requirement from MUST to SHOULD for authors.

scottaohara:
provide note explaining why this is an issue which can be avoided by simply using lowercase.

add additional example/reference to casing requirement to examples of incorrect usage section.

make case requirements a sub-section of “document conformance requirements for use of aria attributes in html” rather than it’s own top-level section.

this PR is a middle ground between keeping the current requirement, and the change proposed in #287

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/333.html" title="Last updated on Jun 28, 2021, 3:35 PM UTC (f1b779a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/333/0c824f8...f1b779a.html" title="Last updated on Jun 28, 2021, 3:35 PM UTC (f1b779a)">Diff</a>

PR #392, opened 2022-01-10
add new aria 1.2 roles

### LINKED ISSUE #200: Add new ARIA roles
The following roles are new in ARIA 1.2 and need to be added to the spec

- [x] `blockquote` => `<blockquote>`
- [x] `caption` => `<caption>` and <s>`<figcaption>`</s>
- [x] `code` => `<code>`
- [x] `definition` => revised, no HTML element match
- [x] `deletion` => `<del>`
- [x] `emphasis` => `<em>`
- [ ] `generic` => `<div>`, `<span>`, & other user facing elements with no meaningful ARIA role.
- [x] `insertion` => `<ins>`
- [x] `meter` => `<meter>`
- [x] `paragraph` => `<p>`
- [x] `strong` => `<strong>`
- [x] `subscript` => `<sub>`
- [x] `superscript` => `<sup>`
- [x] `term` => `<dfn>` and not `<dt>`
- [x] `time` => `<time>`

(removed `label` and `legend` these likely won't be added anymore)

## ARIA 1.3 roles
- `mark` => `<mark>` 
- `suggestion` => no HTML element 
- `comment` => no HTML element 

The following roles may be changed in their naming
- `associationlist` => `<dl>` 
- `associationlistkey` => `<dt>` 
- `associationlistvalue` => `<dd>`

scottaohara:
closes #200

Adds `blockquote`, `caption`, `code`, `deletion`, `emphasis`, `generic`, `insertion`, `meter`, `paragraph`, `strong`, `subscript`, `superscript`, `time`.

Updates links from referencing ARIA 1.1 to ARIA 1.2

changes / corrects `dd` and `dt` to not reference an ARIA role, as `definition` and `term` were corrected in ARIA 1.2 to not be associated with these elements.

These are non-normative updates, since these roles are defined in ARIA 1.2/CORE AAM and the HTML mappings in HTML AAM, respectively.

TODO:
more elements map to `generic`.   <s>Additionally, there needs to be some content added to the spec to indicate that `generic` is a role authors SHOULD NOT be using, per the ARIA spec.  This could be tacked onto the 'any role' definition</s> (issue to update 'any role' #358).

related to resolving HTML AAM https://github.com/w3c/html-aam/issues/373

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/392.html" title="Last updated on Mar 12, 2022, 2:34 PM UTC (42237a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/392/7ccc612...42237a7.html" title="Last updated on Mar 12, 2022, 2:34 PM UTC (42237a7)">Diff</a>


stevefaulkner (approved):

scottaohara:
Waiting on merging of https://github.com/w3c/html-aam/pull/364 before finishing this PR

index.html
```diff
@@ -169,25 +168,27 @@ <h3>
...
+          unnecessarily making the markup more verbose, and incorrectly signaling to other authors 
+          that this practice is useful. Please review the section ["Be cautious of side effects"](#side-effects)
+          for an example of where specifying unnecessary roles can be problematic.
         </p>
         <pre class="example HTML" title="Redundant role on button">
           &lt;!-- Avoid doing this! -->
           &lt;button role="button"&gt;...&lt;/button&gt;
         </pre>
         <p>
           Similarly, the following uses a `role=group` on a [^fieldset^] element, and a `role=Main` on a [^main^] element.
```
patrickhlauke, line 187: lowercase the "M" in `role=Main`

patrickhlauke: or is this done on purpose?

if not, also correct it in the code sample that follows

index.html
```diff
@@ -227,6 +228,34 @@ <h3>
...
       </section>
+      <section class="informative">
+        <h3>Adhere to the rules of ARIA</h3>
+        <p>
+          [[[wai-aria-1.2]]] defines a number of roles which are not meant to be used
+          by authors. Many of these roles are categorized as <a data-cite="wai-aria-1.2#isAbstract">Abstract Roles</a>
+          and are absolutely not to be used by authors. The following example illustrates the invalid use of an
+          abstract `select` role, where an author likely should have used a `combobox` role instead.
+        </p>
+        <pre class="HTML example" title="Abstract roles are not for authors">
```
patrickhlauke, line 239: is there a reason for the `title` here? don't think it's exposed to AT etc, so only mouse users would see/experience this? better to add as a first comment in the markup example instead?

index.html
```diff
@@ -227,6 +228,34 @@ <h3>
...
+          feature parity with a number of HTML elements that do not have more specific ARIA semantics of their
+          own. For instance, HTML's [^div^] and [^span^] elements, among others. ARIA discourages authors from
+          using the `generic` role as its intended purpose is for use by implementors of user agents.
+        </p>
+        <p>
+          In the following example, rather than using a `generic` role, authors are advised to either
+          use a `div` in place of the `article` element. Or, if changing the HTML element is not possible,
+          specify a role of `presentation` or `none` instead.
+        </p>
+        <pre class="HTML example" title="Do not specify elements as generic">
```
patrickhlauke, line 254: ditto

patrickhlauke (changes requested):
Small changes/questions - see comments (not sure why it's now marking them as outdated...unless i've commented on some previous state/github wasn't showing me the latest, not sure)

index.html
patrickhlauke: ah, i see this is processed into an actual visible "Example X: ..." thing in the spec when rendered. disregard.

index.html
scottaohara: later in that paragraph it references the case sensitivity section.

patrickhlauke: feels like it's trying to make two separate points though, oddly mashed together.  it starts off with "these two have roles, but they're redundant because in practice it's the same as long as they're lowercase". but one isn't. but it then doesn't explain what the deal with that second one not being lowercase is.

best to keep the two points separate, or explain them properly.

patrickhlauke (approved):
Happy for this to be merged, but worth opening a separate PR to have another look at that `role=Main` issue (as if it's making a point about lowercase, it buries the lede/doesn't actually explain it)

index.html
scottaohara: this wasn't introduced with this PR, but rather https://github.com/w3c/html-aria/pull/333 which you approved :P

scottaohara: i'm bustin on you :)

you want to remove or revise it?

patrickhlauke: "bustin' makes me feel good"

nah it's ok, will revisit that in another PR, as commented. this PR per se is fine

scottaohara:
The only thing that will be out of sync with merging this PR is the `dt` and `dl` mappings in comparison with HTML AAM.  https://github.com/w3c/html-aam/issues/373 will cover that.   Since only those 2 elements won't be accurate until that AAM pr is merged, I think it's probably fine to pull this in now.  I'm going to create a new issue regarding the other elements which still need updates per the TODO in the op.

PR #410, opened 2022-03-27
Clarify li element role allowances

### LINKED ISSUE #351: Be more percise with allowed roles on the li element
The allowed roles on list items (`<li>` elements) are missing a requirement that the parent `ol`, `ul`, `menu` role must be overwritten to allow for these specific roles that would not be valid within a `role=list`.

Furthermore, if a parent list element's is set to `role=none`, then i could see an argument that the allowed roles for a list item be expanded, and author guidance be provided that the list items need to have their roles overwritten, since without the parent role, the list items would be announced, at least in some cases - as memory serves, as disparate single item lists.   Something to mull over, @stevefaulkner @patrickhlauke 

Edit: tested the `ul role=none` with stray `<li>` elements within today.  at least narrator, jaws and nvda with chromium / gecko browsers do not expose lists/list items for these individual items anymore. A series of 3 list items, with a role=none parent, can instead be strung together as a single text string announcement.  Tldr; not as bad as my memory of how these used to behave, but still kind of odd.  In this case, again, it seems worth while to open up the allowances for list items when the parent list is no longer exposed as such.

scottaohara:
Closes #351

This change is to clarify that if an `li` is a child of a `ul`, `ol`, or `menu` element, and those elements are exposed as a `list`, then no other role can be specified on the `li`.

However, if the parent of the `li` is no longer exposed as a `list`, or if someone were to write invalid markup and use an `li` outside of the expected list > listitem markup pattern, then _any_ role should be allowed on the `li` (so long as the role is a role allowed by the `li`'s parent element) 

For instance,
```
<ul role=table>
  <li role=button>...</li>
</ul>
``` 
is not be allowed as a `button` cannot be a direct child of a `table`.  However, but a `role=row` on the `li` would be allowed in this example.

and 
```
<div>
  <li role=paragraph>...</li>
</div>
```
is invalid HTML, but as the `li` is not a child of a list element, it should be allowed to have any role.

[test cases](https://w3c.github.io/html-aria/tests/li-element-roles.html)

- [x] [HTML Validator](https://github.com/validator/validator/issues/1368)
- [x] [IBM Equal Access Accessibility Checker](https://github.com/IBMa/equal-access/issues/824)
- [ ] [axe-Core](https://github.com/dequelabs/axe-core/issues/3439)
- [x] [ARC Toolkit](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/issues/64)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/410.html" title="Last updated on May 31, 2023, 2:16 PM UTC (0b9b50c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/410/4d9fc05...0b9b50c.html" title="Last updated on May 31, 2023, 2:16 PM UTC (0b9b50c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1908,20 +1908,17 @@
               <code>role=<a href="#index-aria-listitem">listitem</a></code>
             </td>
             <td>
-              <p>
-                Roles:
-                <a href="#index-aria-menuitem">`menuitem`</a>,
-                <a href="#index-aria-menuitemcheckbox">`menuitemcheckbox`</a>,
-                <a href="#index-aria-menuitemradio">`menuitemradio`</a>,
-                <a href="#index-aria-option">`option`</a>,
-                <a href="#index-aria-none">`none`</a>,
-                <a href="#index-aria-presentation">`presentation`</a>,
-                <a href="#index-aria-radio">`radio`</a>,
-                <a href="#index-aria-separator">`separator`</a>,
-                <a href="#index-aria-tab">`tab`</a>
-                or <a href="#index-aria-treeitem">`treeitem`</a>
-              </p>
-              <p>
+              <div class="proposed correction">
+                <p>
+                  If the `li` is not a child of a list element (`ul`, `ol`, `menu`), or the 
+                  list element's implicit `list` role has been overwritten, then 
+                  <a><strong>any `role`</strong></a> that is an allowed child of the parent element.
+                </p>
+                <p>
+                  If the `li` is a child of a `ul`, `ol`, or `menu` with no explicit `role`: 
+                  <strong class="nosupport"><a>No `role`</a></strong>.
+                </p>
+              </div>
               <p>
                 <a data-cite="wai-aria-1.2#global_states">Global `aria-*` attributes</a> 
                 and any `aria-*` attributes applicable to the allowed roles.
--- a/index.html
+++ b/index.html
@@ -1910,13 +1910,13 @@
             <td>
               <div class="proposed correction">
                 <p>
-                  If the `li` is not a child of a list element (`ul`, `ol`, `menu`), or the 
-                  list element's implicit `list` role has been overwritten, then 
-                  <a><strong>any `role`</strong></a> that is an allowed child of the parent element.
+                  If the `li` is not a child of a list element 
+                  (<a href="#el-ul">`ul`</a>, <a href="#el-ol">`ol`</a>, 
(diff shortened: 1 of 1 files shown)


scottaohara:
cc @stevefaulkner @patrickhlauke please review

stevefaulkner:
> those elements are exposed as a list, then no other role can be specified on the `li`.

what about 
`<li role=separator>` or `<li role=none>`

mitchellevan:
> what about `<li role=separator>` or `<li role=none>`

I support allowing `role="none"` on `<li>`. have seen this in the wild, and I don't think we should discourage it:

```
<ul>
  <li>Bats</li>
  <li class="purely-decorative" role="none"></li> <!-- this list item has no content -->
  <li>Cats</li>
</ul>
```

I lean against allowing `role="separator"` on `<li>`. Disallowing `role="separator"` would encourage this reasonable structure:

```
<ul>
  <li>Bats</li>
  <li><span role="separator"></span></li>
  <li>Cats</li>
</ul>
```

mitchellevan:
Or if we want the rules to be more opinionated (in the software sense), say one of the following:

* Allow `role="none"` on empty `<li>`, disallow otherwise
* Say `role="none"` SHOULD NOT be on `<li>` (but valid reasons may exist)

scottaohara:
that seems reasonable on its surface, but would make the implementation of the rule more complicated. practically one could do `aria-hidden=true` on the `li` to achieve the same effect, and not have to have additional checks to the rule to determine not only if it was empty, but if it did have children that they only equated to white space, or themselves had alt="", or aria-hidden=true or role=none/presentation, as applicable to the element and its children, if any.

re:
>I lean against allowing role="separator" on `<li>`.

if that were actually allowed by ARIA/HTML, then it would be a reasonable allowance.  Separately I have tested the current reality of how that works with AT, and while it causes some minor breakages depending on the browser/at combo, I have requested that become an allowance in HTML, which would then make it allowed in ARIA, and subsequently this spec.  Until then though, there will be no change here on that matter.

mitchellevan:
> that seems reasonable on its surface, but would make the implementation of the rule more complicated. practically one could do `aria-hidden=true` on the `li` to achieve the same effect

I'm convinced. Thanks for the clear explanation.

patrickhlauke (approved):

PR #434, opened 2022-11-02
Editorial: clarify summary element role mapping

### LINKED ISSUE #355: Implicit ARIA semantics for summary should not be "button"
In the column "Implicit ARIA semantics", summary has "button", but summary has no implicit ARIA semantics

See: https://www.w3.org/TR/html-aam-1.0/#summary-element, https://github.com/whatwg/html/issues/2272, https://github.com/w3c/html-aam/issues/345

scottaohara:
Closes #355

Since the original issue was filed, and I have spent more time researching details/summary and trying to get consensus on what to do with the summary element in particular, I think it does make sense to adjust this per JAWS-test's issue.  The role is now stated as "no corresponding role" since there is no direct ARIA mapping in HTML AAM, but the note was added to indicate the reality of how the role is actually mapped.

This PR contains no normative changes, as element roles are not defined by this specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/434.html" title="Last updated on Nov 2, 2022, 2:52 PM UTC (5cdb5f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/434/29e8e29...5cdb5f2.html" title="Last updated on Nov 2, 2022, 2:52 PM UTC (5cdb5f2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2779,7 +2779,7 @@
                 <a>No corresponding role</a>
               </p>
               <div class="note">
-                Many, but not all, user agents map the `summary` element to the <code>role=<a href="#index-aria-button">button</a></code>
+                Many, but not all, user agents expose the `summary` element with an implicit ARIA <code>role=<a href="#index-aria-button">button</a></code>
                 role.
               </div>
             </td>
--- a/index.html
+++ b/index.html
@@ -2775,7 +2775,13 @@
               [^summary^]
             </th>
             <td>
-              <code>role=<a href="#index-aria-button">button</a></code>
+              <p>
+                <a>No corresponding role</a>
+              </p>
+              <div class="note">
+                Many, but not all, user agents map the `summary` element to the <code>role=<a href="#index-aria-button">button</a></code>
+                role.
+              </div>
             </td>
             <td>
               <p>


index.html
```diff
@@ -2775,7 +2775,13 @@ <h2 id="docconformance">
               [^summary^]
             </th>
             <td>
-              <code>role=<a href="#index-aria-button">button</a></code>
+              <p>
+                <a>No corresponding role</a>
+              </p>
+              <div class="note">
+                Many, but not all, user agents map the `summary` element to the <code>role=<a href="#index-aria-button">button</a></code>
```
```suggestion
                Many, but not all, user agents expose the `summary` element with an implicit ARIA <code>role=<a href="#index-aria-button">button</a></code>
```

patrickhlauke (approved):
Seems ok to me, considering that HTML spec/user agents seem unwilling to change to disallow interactive content inside `<summary>`

stevefaulkner (approved):

JAWS-test:
@scottaohara 

However, the permitted attributes still state: 

> ...  and any aria-* attributes applicable to the button role. 

I suggest removing that as well, since the attributes are not meaningful (e.g. aria-haspopup) or contradict implicit attributes of the HTML element (aria-expanded)

scottaohara:
you're right @JAWS-test.  but that will need to be considered a normative change.  so we need to do that separate.

PR #436, opened 2022-11-04
Explicitly state requirements in 3rd column of conformance table

sideshowbarker:
For all elements that have implicit roles, this change makes the third column of the “Rules of ARIA attribute usage by HTML element” table explicitly state which role values are not recommended or should not be used — that is, the change makes that column explicitly state, for example, “No role other than `presentation`, which is NOT RECOMMENDED”, or “Any role, though `figure` is NOT RECOMMENDED, or “`img` is also allowed, but NOT RECOMMENDED”.

***

### :boom: Error: 504 Gateway Timeout :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 4, 2022, 8:38 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fhtml-aria%2F1ce2245fe950ca4793993b65e35e771232e4d5fb%2Findex.html%3FisPreview%3Dtrue)

```
<html><body><h1>504 Gateway Time-out</h1>
The server didn't respond in time.
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/html-aria%23436.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -411,7 +411,7 @@
                 <a href="#index-aria-radio">`radio`</a>,
                 <a href="#index-aria-switch">`switch`</a>,
                 <a href="#index-aria-tab">`tab`</a>
-                or <a href="#index-aria-treeitem">`treeitem`</a>
+                or <a href="#index-aria-treeitem">`treeitem`</a>. (<code><a href="#index-aria-link">link</a></code> is also allowed, but NOT RECOMMENDED.)
               </p>
               <p>
                 DPub Roles:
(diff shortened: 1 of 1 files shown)


stevefaulkner (approved):

scottaohara:
thanks @sideshowbarker 

unfortunately, i was in the middle of working on fixing exactly this per the validator issue where this topic had come up.

But this is a good stop gap until I finish the updates I was doing.

PR #464, opened 2023-05-11
Clarification: list additional aria attributes for file and color inputs

### LINKED ISSUE #457: Provide a way to set aria-invalid on a file input element
For an `<input type="file">`, there are many scenarios that would result in an invalid file chosen, whether the file type is incorrect or the file size is too large, or there was an issue with the file itself. 

The [ARIA in HTML doc indicates that file input types have no role](https://www.w3.org/TR/html-aria/#el-input-file). In ARIA 1.2, [aria-invalid is no longer a global attribute](https://www.w3.org/TR/wai-aria-1.2/#aria-invalid). This makes it so that it's not possible to apply the invalid state to a file input type. 

It would be great to either revert `aria-invalid` back to a global property, or to add a file input type role and add it to the list of roles that `aria-invalid` is supported on.

scottaohara:
closes #457

`input type=file` can allow for the aria-required, invalid and disabled attributes.  These attributes are already supported by checkers.

Additionally, `input type=color` also allows for the `aria-disabled` attribute.  Note HTML does not allow the required attribute for this element, so aria-required and invalid wouldn't be applicable here.

The HTML Validator, IBM equal access, and axe core all support these attributes now.  So this is safe to merge without need for implementation issues.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/464.html" title="Last updated on May 11, 2023, 2:46 PM UTC (5463e1f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/464/591be7c...5463e1f.html" title="Last updated on May 11, 2023, 2:46 PM UTC (5463e1f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1675,7 +1675,7 @@
                 <a><strong class="nosupport">No `role`</strong></a>
               </p>
               <p>
-                <a data-cite="wai-aria-1.2#global_states">Global `aria-*` attributes</a>.
+                <a data-cite="wai-aria-1.2#global_states">Global `aria-*` attributes</a> and `aria-disabled` attribute.
               </p>
             </td>
           </tr>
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

patrickhlauke:
LGTM

PR #514, opened 2024-04-15
Addition: popovertarget and command guidance

### LINKED ISSUE #513: relationship between native semantics of HTML popover feature and ARIA might need to be normalized.
The current spec hasn't mentioned anything about `popover`, which is a new HTML feature.
The relationship between native semantics of HTML `popover` feature and ARIA might need to be normalized.

scottaohara:
Closes #513

the `aria-expanded` state is implicit when using the `popovertarget` attribute - and authors are not to use the ARIA attribute with this native HTML attribute.

[test case](https://w3c.github.io/html-aria/tests/popover-and-command.html)

- [x] [HTML validator](https://github.com/validator/validator/issues/1855)
- [ ] [IBM equal access accessibility checker](https://github.com/IBMa/equal-access/issues/2334)
- [ ] [axe-core](https://github.com/dequelabs/axe-core/issues/4837)
- [x] [ARC toolkit](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/issues/83)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/514.html" title="Last updated on Dec 1, 2025, 10:04 PM UTC (1b6c215)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/514/bf16983...1b6c215.html" title="Last updated on Dec 1, 2025, 10:04 PM UTC (1b6c215)">Diff</a>


lukewarlow:
This should also be expanded to commandfor once they're in the spec. Doesn't need to be in this PR just leaving a comment so it's noted somewhere.

index.html
```diff
@@ -3641,19 +3645,61 @@ <h3 id="docconformance-attr">
...
+                  This is equivalent to the <a data-cite="html/interaction.html#dom-iscontenteditable">`isContentEditable`</a>
+                  IDL attribute.
+                </p>
+              </th>
+              <td>
+                `aria-expanded="..."`, `aria-details`
+              </td>
+              <td>
+                <p>
+                  Authors MUST NOT use the `aria-expanded` attribute on any element which also has a `command` attribute.
```
lukewarlow: This is possibly too strict? What if you want aria-expanded with your custom command?

lukewarlow:
The table seems to have messed up a bit there's references to contenteditable and then stuff like aria-expanded?

index.html
scottaohara: You’re right. I had intended to list out the command values where that would have been an issue, but only had time to just semi copy the popover rules. 

I’ll make this more nuanced

scottaohara:
> The table seems to have messed up a bit there's references to contenteditable and then stuff like aria-expanded?

Yeah, I’m mid edit on some of this stuff. I’ll get it fixed

PR #498, opened 2023-12-20
Editorial - Improve understandability of "Figure" in "Rules of ARIA attribute usage by HTML element" table

giacomo-petri:
Closes #497 by rewording the "Figure" row in "Rules of ARIA attribute usage by HTML element" table:
- edited if-if-else condition (where else was never applicable)
- re-organized content for a better readability and understandability
- added consistency with other instances of "no rule other than ***, which is not recommended".

---
labels: editorial
---

***
<a href="https://pr-preview.s3.amazonaws.com/giacomo-petri/html-aria/pull/498.html" title="Last updated on Mar 26, 2025, 2:12 PM UTC (744469b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/498/1cf4fc1...giacomo-petri:744469b.html" title="Last updated on Mar 26, 2025, 2:12 PM UTC (744469b)">Diff</a>

PR #548, opened 2025-03-26
Editorial: update figure allowed roles cell

### LINKED ISSUE #497: Improve understandability of Figure in "Rules of ARIA attribute usage by HTML element" table
- **URL**: https://www.w3.org/TR/html-aria/#el-figure
- **Issue description**: 
    The existing content in the "ARIA role, state, and property allowances" cell for the Figure is as follows:
    
    > If the figure has no figcaption descendant:
    > [Any role](https://www.w3.org/TR/html-aria/#dfn-any-role), though [figure](https://www.w3.org/TR/html-aria/#index-aria-figure) is NOT RECOMMENDED.
    > 
    > If the figure has a figcaption descendant:
    > DPub Role: [doc-example](https://www.w3.org/TR/dpub-aria-1.0/#doc-example).
    > 
    > Otherwise, [figure](https://www.w3.org/TR/html-aria/#index-aria-figure) is allowed, but NOT RECOMMENDED.
    > 
    > [Global aria-* attributes](https://www.w3.org/TR/wai-aria-1.2/#global_states) and any aria-* attributes applicable to the allowed roles.

    The if-if-else structure is perplexing because the "otherwise" scenario may never be applicable, given that the second "if" statement is the opposite of the first one and covers the entirety of the options (if has no figcaption descendant, if has figcaption descendant, otherwise).

    Additionally, the "Otherwise" condition diverges from the typical instances found in the document, where it generally advises "No role other than ***, which is NOT RECOMMENDED.".

    I am submitting a pull request that introduces a suggested resolution to address these concerns.

scottaohara:
closes #497.  supersedes #498

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/548.html" title="Last updated on Mar 26, 2025, 2:46 PM UTC (3ed8d19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/548/1cf4fc1...3ed8d19.html" title="Last updated on Mar 26, 2025, 2:46 PM UTC (3ed8d19)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1295,7 +1295,7 @@
                 <a data-cite="dpub-aria-1.0#doc-example">`doc-example`</a>.
               </p>
               <p>
-                If the `figure` has no `figcaption` descendant:
+                Otherwise, if the `figure` has no `figcaption` descendant:
                 <br>
                 <a><strong>Any `role`</strong></a>, though <code><a href="#index-aria-figure">figure</a></code> is NOT RECOMMENDED.
               </p>
--- a/index.html
+++ b/index.html
@@ -1285,18 +1285,19 @@
             </td>
             <td>
               <p>
+                If the `figure` has a `figcaption` descendant:
+                <br>
+                <a><strong class="nosupport">No `role`</strong></a> other than <code><a href="#index-aria-figure">figure</a></code>, 
+                which is NOT RECOMMENDED.
+              </p>
+              <p>
+                DPub Role:
+                <a data-cite="dpub-aria-1.0#doc-example">`doc-example`</a>.
+              </p>
+              <p>
                 If the `figure` has no `figcaption` descendant:
                 <br>
                 <a><strong>Any `role`</strong></a>, though <code><a href="#index-aria-figure">figure</a></code> is NOT RECOMMENDED.
-              </p>
-              <p class="addition proposed">
-                If the `figure` has a `figcaption` descendant:
-                <br>
-
-                DPub Role:
-                <a data-cite="dpub-aria-1.0#doc-example">`doc-example`</a>.</p>
-
-                <p>Otherwise, <code><a href="#index-aria-figure">figure</a></code> is allowed, but NOT RECOMMENDED.
               </p>
               <p>
                 <a data-cite="wai-aria-1.2#global_states">Global `aria-*` attributes</a> 
--- a/index.html
+++ b/index.html
@@ -1285,7 +1285,7 @@
             </td>
             <td>
               <p>
-                If the `figure` has a `figcaption` descendant:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1285,18 +1285,19 @@ <h2 id="docconformance">
             </td>
             <td>
               <p>
-                If the `figure` has no `figcaption` descendant:
-                <br>
-                <a><strong>Any `role`</strong></a>, though <code><a href="#index-aria-figure">figure</a></code> is NOT RECOMMENDED.
-              </p>
-              <p class="addition proposed">
                 If the `figure` has a `figcaption` descendant:
```
```suggestion
                If the `figure` has a valid `figcaption` descendant:
```

index.html
```diff
@@ -1285,18 +1285,19 @@ <h2 id="docconformance">
...
-
+              <p>
                 DPub Role:
-                <a data-cite="dpub-aria-1.0#doc-example">`doc-example`</a>.</p>
-
-                <p>Otherwise, <code><a href="#index-aria-figure">figure</a></code> is allowed, but NOT RECOMMENDED.
+                <a data-cite="dpub-aria-1.0#doc-example">`doc-example`</a>.
+              </p>
+              <p>
+                If the `figure` has no `figcaption` descendant:
```
```suggestion
                Otherwise, if the `figure` has no `figcaption` descendant:
```

scottaohara:
@giacomo-petri sorry for the long delay in responding to your prior PR.  

I've made some adjustments to what you were suggesting and made this new PR.  Please take a look when you can to verify if this restructuring still accomplishes what you had hoped.

giacomo-petri:
It works, thanks!

Minor tweak: I don't think there are many cases like this, but I'm wondering if there is a simpler way to clarify that the DPUB condition is nested within the if condition.

For example:
```
If the figure has a valid figcaption descendant:
<ul>
  <li>No role other than ...</li>
  <li>DPUB role: doc-example</li>
</ul>

Otherwise...
```

I realize that this might require significant effort for consistency with the whole document, so I'm wondering if there's a simpler solution.

That said, it's already clear enough and resolves my main concern with the "if true - if false - else" structure, where the else condition was never applicable. If we come up with a better approach in the future, we can apply it then.

giacomo-petri (approved):

scottaohara:
thanks @giacomo-petri 

yeah, i was thinking about similar - but deferred to just the current structure because it would mean a larger effort to make other instances consistent would need to be made.  and per that, i'd much rather do further restructuring to the table (as we started talking about in our last phone conversation) to alleviate such issues, than do a bunch of work and _then_ do other structural changes that would undo that.

so i'm going to leave it for now, but note that your point is understood and valid.  thank you.

PR #550, opened 2025-04-01
Correction for HTML element role mapping

scottaohara:
per https://github.com/w3c/aria/pull/2504 - the html element doesn't actually map to the `document` role - rather a parent `document node` to the `html` element is what exposes the `document` role of a web page.

This update fixes the spec to match the updated HTML aam mapping / matches reality.  

The `generic` role is being added as an "allowed" but not recommended role to the `html` element.  The `document` role is being kept as an allowed role for the `html` element.  The rational there being that this has been in the spec for a long time and even when specified on the html element, from what i can tell, there's no pro/con to its inclusion.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/550.html" title="Last updated on Jul 23, 2025, 5:09 PM UTC (7abe723)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/550/a4ba0a6...7abe723.html" title="Last updated on Jul 23, 2025, 5:09 PM UTC (7abe723)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -65,6 +65,11 @@
       </p>
       <ul>
         <li>
+          <a href="https://github.com/w3c/html-aria/pull/550">23 July 2025 - Correction:</a>
+          Clarify that the <a href="#el-html">`html`</a> element is a `generic` element, and that
+          neither the `document` or `generic` roles are recommended to be used on the element.
+        </li>
+        <li>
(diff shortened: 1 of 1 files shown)


pkra (approved):

scottaohara:
this will merge after https://github.com/w3c/aria/pull/2504 is accepted

PR #556, opened 2025-07-23
Addition: update allowances for label element

### LINKED ISSUE #492: Revise allowed roles for label element
the rules around what roles are allowed on label should better refelct when adjusting the role will cause an accessibility issue. 

there are really 3 differnet use cases:

1. label is associated with a form element via its for attribute
2. label is associated with a form element by being its ancestor
3. label is not associated with a form element

For 1 & 2, there can be accName breakages if the role is changed - and for 2 specifically there can be potential invalid nesting of role issues we would want to avoid

For 3.... someone decided to use `label` instead of `span`.  For most cases, I'd submit there's not a whole lot that can go wrong here if someone were to modify the role so the element could be exposed as / behave as something else.

scottaohara:
Closes #492

This update keeps the same allowances for when a `label` element is associated with a labelable element.  But for instances where an author just uses the `label` element without associating it with a labelable element, then the ARIA allowances should replicate the `span` element.

[test case](https://w3c.github.io/html-aria/tests/label-element.html)

- [x] [HTML validator](https://github.com/validator/validator/issues/1856)
- [ ] [IBM equal access accessibility checker](https://github.com/IBMa/equal-access/issues/2335)
- [ ] [axe-core](https://github.com/dequelabs/axe-core/issues/4838)
- [x] [ARC toolkit](https://github.com/ThePacielloGroup/WAI-ARIA-Usage/issues/84)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/556.html" title="Last updated on Jul 23, 2025, 6:55 PM UTC (fdf02e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/556/a497ae5...fdf02e2.html" title="Last updated on Jul 23, 2025, 6:55 PM UTC (fdf02e2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2241,7 +2241,7 @@
                 <a><strong class="nosupport">no `role`</strong></a>
               </p>
               <p>
-                Otherwise, if the `label` is not associted with an element then <a><strong>any `role`</strong></a> 
+                Otherwise, if the `label` is not associted with an element then <a><strong>any `role`</strong></a>, 
                 though <code><a href="#index-aria-generic">generic</a></code> SHOULD NOT be used.
               </p>
               <p><a>Naming Prohibited</a> if exposed as the `generic` role, or if exposed as another role which prohibits naming.</p>
--- a/index.html
+++ b/index.html
@@ -2236,9 +2236,15 @@
             </td>
             <td>
               <p>
-                <a><strong class="nosupport">No `role`</strong></a>
+                If a `label` element is implicitly or explicitly associated with a 
+                <a href="data-html/https://html.spec.whatwg.org/multipage/forms.html#category-label">labelable element</a> then
+                <a><strong class="nosupport">no `role`</strong></a>
               </p>
-              <p class="addition"><a>Naming Prohibited</a></p>
+              <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2236,9 +2236,15 @@ <h2 id="docconformance">
...
             <td>
               <p>
-                <a><strong class="nosupport">No `role`</strong></a>
+                If a `label` element is implicitly or explicitly associated with a 
+                <a href="data-html/https://html.spec.whatwg.org/multipage/forms.html#category-label">labelable element</a> then
+                <a><strong class="nosupport">no `role`</strong></a>
               </p>
-              <p class="addition"><a>Naming Prohibited</a></p>
+              <p>
+                Otherwise, if the `label` is not associted with an element then <a><strong>any `role`</strong></a>
```
patrickhlauke:
```suggestion
                Otherwise, if the `label` is not associted with an element then <a><strong>any `role`</strong></a>, 
```

missing comma

patrickhlauke (approved):

PR #578, opened 2026-04-15
informative li element role clarification

### LINKED ISSUE #577: clarify li in list element with non-list role
Motivated by https://github.com/dequelabs/axe-core/issues/5073#issuecomment-4251293900

From https://w3c.github.io/html-aria/#el-li

> If the li is a child of a list element ([ul](https://w3c.github.io/html-aria/#el-ul), [ol](https://w3c.github.io/html-aria/#el-ol), [menu](https://w3c.github.io/html-aria/#el-menu)) role=[listitem](https://w3c.github.io/html-aria/#index-aria-listitem).

This is confusing as it only mentions the list element and does not consider if authors have given the list element a (non-list) role - in which case the LI should be generic (as per HTML-AAM).

scottaohara:
closes #577

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/578.html" title="Last updated on Apr 15, 2026, 12:35 PM UTC (70258b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/578/d29fd9f...70258b5.html" title="Last updated on Apr 15, 2026, 12:35 PM UTC (70258b5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2292,14 +2292,14 @@
             </th>
             <td>
               <p>
-                If the `li` is a child of a list element 
-                (<a href="#el-ul">`ul`</a>, <a href="#el-ol">`ol`</a>, 
-                <a href="#el-menu">`menu`</a>)
+                If the `li` is a child of an element exposed as a `role=list` 
+                (e.g., <a href="#el-ul">`ul`</a>, <a href="#el-ol">`ol`</a>, 
+                <a href="#el-menu">`menu`</a>):
                 <code>role=<a href="#index-aria-listitem">listitem</a></code>.
               </p>
               <p>
-                Otherwise, if the `li` is not a child of a list element it is exposed as 
-                a <code>role=<a href="#index-aria-generic">generic</a></code>.
+                Otherwise, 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2292,14 +2292,14 @@ <h2 id="docconformance">
             </th>
             <td>
               <p>
-                If the `li` is a child of a list element 
-                (<a href="#el-ul">`ul`</a>, <a href="#el-ol">`ol`</a>, 
-                <a href="#el-menu">`menu`</a>)
+                If the `li` is a child of an element exposed as a `role=list`
```
pkra, line 2295: I thought this should keep "list element" but it seems `<div role="list"> <li>item</li> </div>` keeps `li` as a listitem role across the board - wow.

pkra (approved):

index.html
scottaohara: yup.  also, wanted to make it clearer that if the list element's role is overwritten, then it's no longer a `list` - so it's not just based on if the parent is a list "element".

PR #585, opened 2026-05-05
Fix role attribute case in main element

scottaohara:
closes #583

lowercases the "m" to the role of `main` in the "Redundant role on fieldset and main" example

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/pull/585.html" title="Last updated on May 19, 2026, 6:46 PM UTC (f0ca33e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aria/585/d7ea2ba...f0ca33e.html" title="Last updated on May 19, 2026, 6:46 PM UTC (f0ca33e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -349,16 +349,14 @@
           &lt;button role="button"&gt;...&lt;/button&gt;
         </pre>
         <p>
-          Similarly, the following uses a `role=group` on a [^fieldset^] element, and a `role=Main` on a [^main^] element.
+          Similarly, the following uses a `role=group` on a [^fieldset^] element, and a `role=main` on a [^main^] element.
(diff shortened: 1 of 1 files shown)


pkra (approved):

index.html
jyasskin: Note that this sentence appears to be here specifically because of the incorrect capitalization in the example. I'm not sure why the original authors included an "as long as [false]" clause, but this might need more investigation.

cc/ @shirsha

jyasskin: The casing here should match the content of the example, but see my other comment for a question on whether this is implying something that's not obvious.

```suggestion
          Similarly, the following uses a `role=group` on a [^fieldset^] element, and a `role=main` on a [^main^] element.
```

